PR #49, opened 2015-05-14
(Ready for Review) API for permissions and library to handle logic

lmanotas:


candyj0311:
This should be closed out.

PR #58, opened 2016-01-27
Remove Ortuna from pdr_dev.sql

rickymetz:
Removed the Ortuna lines in `pdr_dev.sql` which was causing database to crash.


rickymetz:
@bithavoc @seanperkins

bithavoc:
LGTM

PR #66, opened 2016-01-29
Client and Server Source Code unified

bithavoc:
Merge [client](https://github.com/MobilityLabs/pdredesign-client) source code


MakotoTheKnight:
We will definitely need a separate task to go through Hakiri and make it happy.  This specifically, which has been around for a while.

https://hakiri.io/github/MobilityLabs/pdredesign-server/master/848d088ea94726ca4676584b894e677df7b93cb6/warnings?name=Session+Setting

bithavoc:
Yes, I'll create it.

MakotoTheKnight:
The only really major thing I'm noticing here is that the `.ruby-version` file is out of sync with the Gemfile.  It does make things very inconvenient while working in RBEnv; I'd say bump the version in that file to 2.2.3 to match the Gemfile.

Overall it looks fine but I still plan to meet up with you tomorrow to talk about it.

MakotoTheKnight:
This looks good to me.  Talked with @bithavoc on this PR and we agreed that the merging of the two repositories seems sound.

There may be some minor hiccups, but we don't anticipate that there will be anything showstopping.   We want to get this deployed into staging and on master so that we could smoke test it.

PR #72, opened 2016-02-03
Closes #60:  Update modals to display only facilitator and participant

### LINKED ISSUE #60: Update modals to display only facilitator and participant

MakotoTheKnight:

 - The modal from direction received by @seanperkins was indicated in the two spots shown in screenshots.  This is after an assessment is created and one wishes to see the dashboard.
 - If the user already has the "viewer" role, we will display the value to not break the view, but we will not allow them to reselect it in the future.
 - This removes the Codeclimate chunks of test code in the repository.  We've disabled the service and if we really want it back, we can always resurrect it later.
 - The production-facing code changes were relatively light in this one; the main lift comes in from test hygiene and cleaning up some expectations, as well as ensuring that the tests pass.

image

image

### DIFF
--- a/lib/assessments/permission.rb
+++ b/lib/assessments/permission.rb
@@ -46,12 +46,12 @@ module Assessments
           return false
       end
       notify_user_for_access_granted(assessment, user, level)
-      return true
+      true
     end
 
     def update_level(user, level)
       unless assessment.owner?(user)
-        unless(get_level(user).to_s == level.to_s)
+        unless get_level(user).to_s == level.to_s
           revoke_level(user)
           add_level(user, level)
         end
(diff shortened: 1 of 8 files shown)


lib/assessments/permission.rb
```diff
@@ -73,23 +77,23 @@ def revoke_level(user)
           assessment.viewers.destroy(user)
         when :network_partner
           assessment.network_partners.destroy(user)
-        assessment.reload
+          assessment.reload
```
MakotoTheKnight: This statement's giving me pause - I'm going to be sure that this didn't cause any regressions I didn't intend to!

seanperkins: Do you mean that is looks like it was supposed to reload inside `update_level` either way?

MakotoTheKnight: I originally thought that, but it seems that the original statement had it encapsulated inside of the last `when` block.  I'm going to triple check that this is the case; RubyMine can be pretty zealous about code formatting.

MakotoTheKnight: Okay - quick smoke test of general Ruby switch statements indicates that we're okay here.  Conventionally you only see one line of action inside of a `when` which is what gave me pause to begin with, but I believe that we're alright.

bithavoc:
LGTM

PR #73, opened 2016-02-03
Closes #61:  Update invitations to only include facilitator and participant

### LINKED ISSUE #61: Update invitations to only include facilitator and participant

MakotoTheKnight:

 - Most of the work to do this effort was already covered in #72, which is why this branch is based off of that.  This PR will need to go after it.
 - The only major change required in this task was to remove the pop-over text which described what a viewer is.  It would be present in #72 and master, but not here.


bithavoc:
LGTM

PR #76, opened 2016-02-15
Closes #71:  Use FactoryGirl

### LINKED ISSUE #71: Use FactoryGirl with traits instead of globally defined instance variables for tests
Just a bit of test hygiene; we shouldn't be using globally defined variables for our testing efforts.  That should be relegated to FactoryGirl instead.

MakotoTheKnight:

 - This attempts to remove global state throughout the test suite and force tests to specify their own state when testing.
 - Many conventions have been preserved insofar as what object they're creating, for instance:  `FactoryGirl.create(:user, :with_district)` would be the conventional way to create a user with a predefined district, for example.
 - No production code has been touched for this exercise; all tests should pass so long as we substitute FactoryGirl for global state.

The goal is to replace global state in our tests as we progress forward with our work.


MakotoTheKnight:
@bithavoc Could I get you to take a look into this?  I wouldn't want us to get too far without this backing.

spec/authorizers/organization_authorizer_spec.rb
```diff
@@ -1,23 +1,24 @@
 require 'spec_helper'
 
 describe OrganizationAuthorizer do
-  before do
-    @org = Organization.create!(name: 'test')
-  end
+  let(:org) {
+    Organization.create!(name: 'test')
```
bithavoc: any reason why we're not using a factory for the organization too?

spec/factories/districts.rb
```diff
@@ -0,0 +1,43 @@
...
+#  metmic        :string(255)
+#  ulocal        :string(255)
+#  cdcode        :string(255)
+#  bound         :string(255)
+#  created_at    :datetime
+#  updated_at    :datetime
+#
+
+FactoryGirl.define do
+  factory :district do
```
bithavoc: did we forgot to add attributes to this factory?

spec/authorizers/organization_authorizer_spec.rb
MakotoTheKnight: No time.  The main goal was to attack `Application::create_sample_user` and its kind.

spec/factories/districts.rb
MakotoTheKnight: No; none of the fields are required explicitly before we can create a district.  We also didn't have any tests that depended on any metadata to be contained in them before we newed them up or used them in the test, so I didn't see it appropriate to add anything in there.

spec/mailers/access_granted_mailer_spec.rb
```diff
@@ -1,15 +1,32 @@
...
-  let(:user){ Application.create_user } 
-  let(:role){ "facilitator" }
-
-  it "#notify" do
-    mail = subject.notify(assessment, user, role)
-    expect(mail.to).to include(user.email)
-    expect(mail.body).to include(role)
+  let(:subject) {
+    AccessGrantedMailer
+
```
bithavoc: should I ignore these blank spaces in the code review? rubyists don't like them

MakotoTheKnight: That's something we should tackle with a tool that helps us with formatting on commit, but I would say if we want to address it that we can address it when we come across it again.  It doesn't impact the tests.

spec/factories/districts.rb
bithavoc: and we're leaving an empty block because it's required by factorygirl?

MakotoTheKnight: I've never seen a `factory` statement without a block (even if it's empty), and I'm not entirely sure if it's supported without a block, but that would be interesting.  By and large, this is me following the prescribed conventions and examples to creating factories.

bithavoc: can you send me a link to the prescribed convention so I can read more?

bithavoc:
LGTM :rocket:

spec/factories/districts.rb
MakotoTheKnight: There's really no link that I can show you; this is more or less what I've read online from any guide describing a factory for FactoryGirl.  It can work just fine without a block, so I've removed it.  We can progress from there.

PR #92, opened 2016-02-17
V2 style initial

rickymetz:


MakotoTheKnight:
It looks like we stomped out a class called `scored-skipped` which controlled the color of the skipped message.  The commit that did that was [this one](https://github.com/MobilityLabs/pdredesign-server/commit/7ef39ffcae4ace72a22f856aa4a2babc53eadd55).

MakotoTheKnight:
We'll want to recreate this PR so that we're pointing at the right branch.

PR #94, opened 2016-02-19
Closes #75: share report link

### LINKED ISSUE #75: Share report by link
* There will be only one share code per assessment
* A button to share will appear in the right side bar of the report page http://localhost:3000/#/assessments/45/report with a mail:to link to the report so it can be shared via email
* Report will be shown as non-facilitator model, which means, the **EDIT** button will be hidden and Dashboard and Consensus will be also invisible to the viewer user.
* In the navigation bar, only leave The brand logo and Sign-in/Signup action
* Only Facilitators will be able to share the link

bithavoc:


spec/javascripts/controllers/AssessmentSharedReportCtrlSpec.js
```diff
@@ -0,0 +1,12 @@
+describe('Controller: AssessmentSharedReportCtrl', function() {
+  var subject, scope;
+
+  beforeEach(module('PDRClient'));
+
+  beforeEach(inject(function($injector, $controller, $rootScope) {
```
MakotoTheKnight: I'm a little unclear on this test; what was it asserting?

spec/javascripts/directives/sharedTakeAwaySpec.js
```diff
@@ -0,0 +1,40 @@
...
+
+    isolatedScope = element.isolateScope();
+    $httpBackend.flush();
+  }));
+
+    it('sets token correctly', function(){
+      expect(isolatedScope.token).toEqual('token-123');
+    });
+
+    it('isFacilitator returns true if assessment.is_facilitator is true', function(){
```
MakotoTheKnight: Looking at this specific test makes me think that it should be its own description; that is, we describe a state in which `assessment.is_facilitator` is true, then we assert that `isFacilitator` is also true.

spec/javascripts/directives/sharedTakeAwaySpec.js
```diff
@@ -0,0 +1,40 @@
...
+      isolatedScope.assessment.is_facilitator = true;
+      expect(isolatedScope.isFacilitator()).toEqual(true);
+    });
+
+    it('isFacilitator returns true', function(){
+      expect(isolatedScope.isFacilitator()).toEqual(true);
```
MakotoTheKnight: This particular test does the same thing as the above one.  I'm unclear as to what this one was meant to catch since the above test implies that there is a scenario in which `isFacilitator` should return `false`.

config/routes.rb
```diff
@@ -28,6 +28,12 @@
     resources  :categories, only: :index
     resources  :prospective_users, only: :create
     resources  :assessments, only: [:index, :show, :update, :create] do
+      collection do
+        resources :shared, controller: 'shared_assessments', only: [:show], param: :token do
```
MakotoTheKnight: This seems a bit strange to me - we're only using one resource on `shared_assessments` and using it as a namespace for other resources below.

We're not using the routes generated from this anywhere that I can see, but I'd like to see how open you would be to changing this up a bit?

Something like:

```
collection do
  get :show, to: 'shared_assessments#show', as: 'shared_assessments', path: '/shared/:token'
  get :report, to: 'shared_report#show', as: 'shared_reports', path: '/shared/:shared_token/report'
  get :priorities, to: 'shared_priorities#index', as: 'shared_priorities', path: '/shared/:shared_token/priorities'
end
```

This reads a bit more verbose, but the idea is that it provides the same routes.

Thoughts?

spec/controllers/v1/shared_report_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+describe V1::SharedReportController do
+  render_views
+
+  before { create_magic_assessments }
+  before { create_struct }
+  let(:assessment) { @assessment_with_participants }
+  let(:consensu){ assessment.response.nil? ? Response.create(responder_id:   assessment.id, responder_type: 'Assessment') : assessment.response }
+
+  describe '#show' do
+    it 'gets a report' do
```
MakotoTheKnight: Is the implicit test here that we're not authenticated?

app/assets/javascripts/client/views/assessments/shared_report.html
```diff
@@ -0,0 +1,151 @@
+<div class='report'>
```
MakotoTheKnight: There are about three differences between this and the existing report.html.  For now this approach will likely do, but we should consider ways to reuse our view components, since we've got the copy between the two duplicated.

MakotoTheKnight:
Found this:

image

MakotoTheKnight:
So I'm not sure if this is a bug or not, but I did notice an unusual state I could get myself into.

I set up an assessment and scheduled it for 1 April.  Nothing heinous about that.

april_first

I immediately loaded this in the unauthenticated view and received...-41 days ago.

into_the_future

[The math pans out](http://www.wolframalpha.com/input/?i=19+February+2016+-+1+April+2016) in that 1 April is _roughly_ 41 days from now, but the odd thing with this display is, if it's meant to be borrowing from the private view, why does the public view say something completely different to the private view?

I'm not sure if it's a bug since I'm uncertain if the typical use case would cause one to ever reach that view while the assessment is yet incomplete and the consensus meeting has yet to be held, but my gut feeling is that the two views should be saying the same thing here.

config/routes.rb
bithavoc: The fact that we're not using generated routes doesn't give me enough reasons to change this over full paths. It's unclear to me what's the advantage of your proposed solution.

spec/controllers/v1/shared_report_controller_spec.rb
bithavoc: There is no `sign_in`, no reason to do `sign_out`

bithavoc:
It was my understanding that you only were able to share the report once the consensus was held, it seems like I was wrong. I restored the original behavior of the private view so it no longer shows negative days.

bithavoc:
@MakotoTheKnight please give this PR a second pass.

app/assets/javascripts/client/controllers/assessments/AssessmentSharedDashboardSidebarCtrl.js
```diff
@@ -0,0 +1,25 @@
+PDRClient.controller('AssessmentSharedDashboardSidebarCtrl', [
```
MakotoTheKnight: This particular controller is missing some key functionality which controls the display of the dates.

config/routes.rb
MakotoTheKnight: It's more a namespacing thing than anything else.  We're only getting one action out of those controllers, so it doesn't feel like a resource.

It's fine the way it is now, though.

bithavoc: I don't understand what you mean with namespacing.

Don't you think it's a little weird to repeat the prefix `/shared/:shared_token/` in three different paths?

Those repeated prefixes seem sinatra-like to me, not rails-like.

spec/controllers/v1/shared_report_controller_spec.rb
MakotoTheKnight: This is fine then.

config/routes.rb
MakotoTheKnight: Like I said, I'm not bothered with the way it is now, but I'm not necessarily opposed to the repetition.  Since we're using three different controllers, the only thing they have in common is the path which you'd get with `:shared`.

app/assets/javascripts/client/controllers/assessments/AssessmentSharedDashboardSidebarCtrl.js
bithavoc: Yep, just added. want to check again?

MakotoTheKnight:
Everything here looks good to go.  :shipit:

PR #95, opened 2016-02-23
Closes #59: simplify permission models

### LINKED ISSUE #59: Simplify permissions models
* Remove viewer user type, now just share by link
* District users can now be facilitator or participant only
* Network partners can be only facilitator

Should take a look at new requirements for Inventory invitations: https://docs.google.com/document/d/1IWKZAtGon0QuZAXbswjOpc2fWX3xZ9MDmG0qHbivHS4/edit#heading=h.e99matkveeo1

May also want to review this PR to see if we can use it to for permissions https://github.com/MobilityLabs/pdredesign-server/pull/49

bithavoc:
New Environment variable `ROLLBACK_ENV` has been set to `staging` for **readiness-assessment-dev** `production` for **readiness-assessment**

### DIFF
--- a/app/assets/javascripts/client/directives/assessments/assessmentLinks.js
+++ b/app/assets/javascripts/client/directives/assessments/assessmentLinks.js
@@ -20,9 +20,10 @@ PDRClient.directive('assessmentLinks', [
           '$location',
           '$timeout',
           '$state',
+          '$q',
           'AccessRequest',
           'SessionService',
-          function($scope, $modal, $rootScope, $location, $timeout, $state, AccessRequest, SessionService) {
+          function($scope, $modal, $rootScope, $location, $timeout, $state, $q, AccessRequest, SessionService) {
             $scope.isNetworkPartner = SessionService.isNetworkPartner;
 
             $scope.linkIcon = function(type){
@@ -75,12 +76,18 @@ PDRClient.directive('assessmentLinks', [
             };
 
             $scope.requestAccess = function() {
-              $scope.modal = $modal.open({
-                templateUrl: 'client/views/modals/request_access.html',
-                scope: $scope,
-                windowClass: 'request-access-window',
-                opened: $scope.addRequestPopover()
-              });
+              if($scope.isNetworkPartner()) {
+                $scope.performAccessRequest('facilitator').then(function() {
+                  console.log("requested")
+                });
+              } else {
+                $scope.modal = $modal.open({
+                  templateUrl: 'client/views/modals/request_access.html',
+                  scope: $scope,
+                  windowClass: 'request-access-window',
+                  opened: $scope.addRequestPopover()
+                });
+              }
             };
 
             $scope.close = function() {
@@ -93,13 +100,21 @@ PDRClient.directive('assessmentLinks', [
             };
 
             $scope.submitAccessRequest = function(roles) {
+              $scope.performAccessRequest(roles).then(function() {
+                $scope.modal.dismiss('cancel');
+              })
+            };
+
+            $scope.performAccessRequest = function(role) {
+              var deferred = $q.defer();
               AccessRequest
-                .save({assessment_id: $scope.id}, {roles: [roles]})
-                .$promise
-                .then(function() {
-                  $scope.modal.dismiss('cancel');
-                  $state.go($state.$current, null, { reload: true });
-                });
+              .save({assessment_id: $scope.id}, {roles: [role]})
+              .$promise
+              .then(function() {
+                $state.go($state.$current, null, { reload: true });
+                deferred.resolve();
+              }, deferred.reject);
+              return deferred.promise;
             };
 
             $scope.gotoLocation   = function(location) {
--- a/app/assets/javascripts/client/directives/assessments/assessmentLinks.js
+++ b/app/assets/javascripts/client/directives/assessments/assessmentLinks.js
@@ -77,9 +77,7 @@ PDRClient.directive('assessmentLinks', [
 
             $scope.requestAccess = function() {
               if($scope.isNetworkPartner()) {
-                $scope.performAccessRequest('facilitator').then(function() {
-                  console.log("requested")
-                });
+                $scope.performAccessRequest('facilitator').then(function() {});
               } else {
                 $scope.modal = $modal.open({
                   templateUrl: 'client/views/modals/request_access.html',
--- a/app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
+++ b/app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
@@ -60,7 +60,7 @@
             <p class="info">
               <fieldset>
                 <label class="level_label">Current Permission Level: </label>
-                <select class="permission-level" name="level">
+                <select class="permission-level" name="level" ng-disabled="user.role == 'network_partner'">
                   <option value={{user.current_permission_level.role}}>
                     {{user.current_permission_level.human}}
                   </option>
--- a/app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
+++ b/app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
@@ -60,7 +60,7 @@
             <p class="info">
               <fieldset>
                 <label class="level_label">Current Permission Level: </label>
-                <select class="permission-level" name="level" ng-disabled="user.role == 'network_partner'">
+                <select class="permission-level" name="level" ng-disabled="user.role === 'network_partner'">
                   <option value={{user.current_permission_level.role}}>
                     {{user.current_permission_level.human}}
                   </option>
--- a/app/services/invitation/insert_from_invite.rb
+++ b/app/services/invitation/insert_from_invite.rb
@@ -69,6 +69,9 @@ module Invitation
     end
 
     def set_permission
+      if invite.user.network_partner?
+        invite.role = 'facilitator'
+      end
       if invite.role
         ap = Assessments::Permission.new(invite.assessment)
         ap.add_level(invite.user, invite.role)
--- a/app/services/invitation/insert_from_invite.rb
+++ b/app/services/invitation/insert_from_invite.rb
@@ -69,9 +69,7 @@ module Invitation
     end
 
     def set_permission
-      if invite.user.network_partner?
-        invite.role = 'facilitator'
-      end
+      invite.role = 'facilitator' if invite.user.network_partner?
       if invite.role
         ap = Assessments::Permission.new(invite.assessment)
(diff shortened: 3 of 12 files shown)


app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
```diff
@@ -60,7 +60,7 @@
             <p class="info">
               <fieldset>
                 <label class="level_label">Current Permission Level: </label>
-                <select class="permission-level" name="level">
+                <select class="permission-level" name="level" ng-disabled="user.role == 'network_partner'">
```
MakotoTheKnight: Using `===` would be more ideal in this scenario.

app/services/invitation/insert_from_invite.rb
```diff
@@ -69,6 +69,9 @@ def generate_password
     end
 
     def set_permission
+      if invite.user.network_partner?
```
MakotoTheKnight: Maybe:

```invite.role = 'facilitator' if invite.user.network_partner?```

This is more of a style thing than anything else, and I'm fine with this as is.

File: app/assets/javascripts/client/views/assessments/permissions/manage_permission.html (hunk 1 above)
Comment (bithavoc) [inline app/assets/javascripts/client/views/assessments/permissions/manage_permission.html]: So tru'

File: app/services/invitation/insert_from_invite.rb (hunk 2 above)
Comment (bithavoc) [inline app/services/invitation/insert_from_invite.rb]: :+1:

File: spec/services/invitation/insert_from_invite_spec.rb [hunk 3]
```diff
@@ -22,6 +22,34 @@ def existing_user_invite
       email:         @user.email)
   end
 
+  describe '#execute' do
+    context 'inviting network partner as participant' do
+      let!(:existing_user) { FactoryGirl.create(:user, :with_network_partner_role) }
```
MakotoTheKnight: If you're using what you declare in a `let` block later on, you don't need the `!`; `let` is lazy and `let!` is eager.

bithavoc: :+1:

MakotoTheKnight:
The non-reversible migration does cause a little bit of a concern.  I can respect that we don't want to add yet more metadata to the database to determine their state prior, but is it possible that we could get environment-specific rollbacks?

We'd have to pull down the resultant entity IDs per environment and allow us to roll them back at will, and even though there's not much of an impact (I think around 50 IDs total in production), I would feel a little better if we had the option to revert.

bithavoc:
@MakotoTheKnight I'd feel better too, any ideas on how to do the environment-specific rollback?

MakotoTheKnight:
I'd imagine something like this as a framework:

```
  def down
    if Rails.env.development?
      # preserve any dev-specific IDs
    elsif Rails.env.production?
      # preserve any production-specific IDs
    end
  end
```

We'd need to run the query you're using to figure out which IDs to capture here, but that shouldn't be too bad.

bithavoc:
mmm interesting, so the migration would contain those ID's hardcoded per environment?

MakotoTheKnight:
Yes.  I can't exactly see another way around it unless we engineered another table to help with that, which at this junction doesn't seem to be as straightforward as the IDs.

bithavoc:
Good catch @MakotoTheKnight, please check again

MakotoTheKnight:
This works, but it'd be nice to disable the button to send a clearer message.

image

bithavoc:
ahh @MakotoTheKnight we agreed that since you can not change the role of a network partner to any of the other roles then we'd disable the option altogether https://github.com/MobilityLabs/pdredesign-server/issues/59#issuecomment-186389228

app/assets/javascripts/client/directives/assessments/assessmentLinks.js
```diff
@@ -75,12 +76,18 @@ PDRClient.directive('assessmentLinks', [
...
             $scope.requestAccess = function() {
-              $scope.modal = $modal.open({
-                templateUrl: 'client/views/modals/request_access.html',
-                scope: $scope,
-                windowClass: 'request-access-window',
-                opened: $scope.addRequestPopover()
-              });
+              if($scope.isNetworkPartner()) {
+                $scope.performAccessRequest('facilitator').then(function() {
+                  console.log("requested")
```
MakotoTheKnight: This console.log seems strange to me; it does the right behavior (with the `performAccessRequest` function), but the console output doesn't serve us any purpose.

bithavoc: lol :+1:

app/assets/javascripts/client/directives/assessments/assessmentLinks.js
```diff
@@ -93,13 +100,21 @@ PDRClient.directive('assessmentLinks', [
             };
 
             $scope.submitAccessRequest = function(roles) {
+              $scope.performAccessRequest(roles).then(function() {
+                $scope.modal.dismiss('cancel');
+              })
+            };
+
+            $scope.performAccessRequest = function(role) {
```
MakotoTheKnight: Okay - this was a bit more complicated than I wanted to read, but I think it makes sense.

You've removed the `$scope.modal.dismiss` operation from within the `$promise` below so that you can pick and choose if you need the modal to be dismissed or not.  In one context it does need to be dismissed, but in the other it doesn't.

Is that what you had in mind for this particular code block?

app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
```diff
@@ -60,7 +60,7 @@
             <p class="info">
               <fieldset>
                 <label class="level_label">Current Permission Level: </label>
-                <select class="permission-level" name="level">
+                <select class="permission-level" name="level" ng-disabled="user.role === 'network_partner'">
```
MakotoTheKnight: So I take back what I said about the form not being disabled; it very much is.  The problem is that it doesn't _appear_ to be at first glance.

db/migrate/20160223195539_migrate_network_partner_participants_to_facilitators.rb
```diff
@@ -0,0 +1,34 @@
+class MigrateNetworkPartnerParticipantsToFacilitators < ActiveRecord::Migration
```
MakotoTheKnight: Will this adversely impact development if we're unable to roll back?

app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
MakotoTheKnight: In that vein we may be missing something here:  we should sanitize the output list that we get back from the server so that it doesn't have a chance of showing up in the view.

If I remove the `disabled` attribute in my DOM, then I see that the network partner still has other roles.  If we're saying that they should only have the one role, we should sanitize the possible roles that they have.

Observe [permission.rb](https://github.com/MobilityLabs/pdredesign-server/blob/59-permissions/lib/assessments/permission.rb#L12); this is where the JSON to send across the wire is being generated.  We'd want to remove facilitator from this list somehow so we don't have to do anything extra in the view to do so.

app/assets/javascripts/client/directives/assessments/assessmentLinks.js
bithavoc: yes

app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
bithavoc: Nice catch, I initially didn't wanted to just catch the Network Partner requesting access piece(like I did in the PR), I also wanted to make it impossible for a network partner to become a participant, however, I thought it'd be too restrictive internally and decided instead to disable the option in the UI.

What I forgot was to restrict the permissions api to perform the change.

So instead limiting the list of role options for a network partner, I'd rather lock down the permissions API to prevent a network partner from becoming something else than a facilitator even if the user removes `disabled` from the **select HTML** element.

Thoughts?

db/migrate/20160223195539_migrate_network_partner_participants_to_facilitators.rb
bithavoc: I'm not sure what you're asking, we agreed on a rollback strategy for staging and production and you're seeing it here.

Worst case scenario, the network partners will still have access to assessments but as facilitators instead of participants.

MakotoTheKnight: I mean, if I run this migration locally to smoke test things and then go to another branch, would I have lost anything, and/or is it necessary for me to roll back my state?

app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
MakotoTheKnight: So long as the canonical source of truth (the API) indicates that a network partner is no longer explicitly supported, that's fine.

db/migrate/20160223195539_migrate_network_partner_participants_to_facilitators.rb
bithavoc: this is a data migration, it depends whether the **other branch** depends on the sate this migration is modifying(network partner roles in assessments).

app/assets/javascripts/client/views/assessments/permissions/manage_permission.html
bithavoc: So I'll override to `facilitator` on any roles for network_partners when modifying the participant role, would that do the trick?

MakotoTheKnight: Possibly.  Give it a go, test your expectations and we can review it.

bithavoc: Our expectations were met as it is, this is your expectation. would locking down the API so it can not be used to switch a network partner into a participant be enough for you?

MakotoTheKnight: Yes, that would be what I'd like to see.

bithavoc: cool

bithavoc:
@MakotoTheKnight please take another look

MakotoTheKnight:
The route `/v1/assessments/:id/permissions/all_users` still brings back the user with more than one possible role on it.

bithavoc:
@makototheknight: The visual option is disabled, even if you manage to reactivate the `select` element for the option the API will block you.

MakotoTheKnight:
Alright, I think we'll need good to go then. :shipit:

PR #96, opened 2016-02-24
Fix Assessment share token migration

bithavoc:
Migrations are currently failing because in staging we have assessments without district.


MakotoTheKnight:
Do we have the same scenario in production as well?

bithavoc:
Sadly yes:
```
irb(main):001:0> Assessment.where(district_id: nil).count
=> 20
```

MakotoTheKnight:
We're going to want to investigate this in the future, but for now, this migration is good by me.  :shipit:

PR #98, opened 2016-02-24
Closes #77, Closes #78, Closes #79: Learning questions

### LINKED ISSUE #77: Learning Questions:  Database Model
- A learning question belongs to a single user in the context of an assessment
- Learning questions are just text up to 255 characters long

### LINKED ISSUE #78: Learning Questions:  Controller
- CRUD resource
- All users associated with an assessment can see all learning questions
- Users may only edit, update, or delete their own learning question

### LINKED ISSUE #79: Learning Questions:  Sidebar Display
- Sidebar is visible throughout all pages of an assessment
- For facilitators:  creating a new learning assessment appears as Step 4
- A modal will appear for all users on the survey display (once loaded) until that specific user has created a question

Rick will be providing mockups for the above pages.

MakotoTheKnight:
This PR is ready.  Notes:

 - Learning questions are created on step 4 by a district partner.
 - Buttons to save and submit all elements have been moved below Step 4 to preserve the aesthetic of completion.
 - Learning questions are available in a modal on the report and the assessment.
 - Learning questions are _not_ available on the questionnaire due to concerns of how the side bar would be able to support them.


app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
```diff
@@ -8,137 +8,137 @@ PDRClient.controller('AssessmentAssignCtrl', [
...
+
+function AssessmentAssignmentCtrlFunction($scope, $timeout, $anchorScroll, $location, $stateParams, SessionService, Assessment, Participant, Rubric) {
+
+  $scope.id = $stateParams.id;
+  $scope.user = SessionService.getCurrentUser();
+  $scope.participants = Participant.query({assessment_id: $scope.id});
+  $scope.rubrics = Rubric.query();
+  $scope.alerts = [];
+
+  $scope.district = $scope.user.districts[0];
```
bithavoc: It's a bit unclear why the user can have multiple districts and we're always picking the first.

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
```diff
@@ -8,137 +8,137 @@ PDRClient.controller('AssessmentAssignCtrl', [
...
-        $scope.datetime = $('.datetime').datetimepicker({
-          pickTime: false,
+  };
+
+  $scope.success = function (message) {
+    $scope.alerts.push({type: 'success', msg: message});
+    $anchorScroll();
+    $timeout(function () {
+      $scope.alerts.splice(message, 1);
+    }, 10000);
```
bithavoc: Shouldn't we have this number somewhere were we can modify globally the constants related to _alerts_?

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
```diff
@@ -8,137 +8,137 @@ PDRClient.controller('AssessmentAssignCtrl', [
...
-          .$promise
-          .then(function(_data) {
-            $scope.saving = false;
-            $scope.success('Assessment Saved!');
-          }, function(){
-            $scope.saving = false;
-            $scope.error('Could not save assessment');
+  AssessmentAssignmentCtrlFunction]);
+
+function AssessmentAssignmentCtrlFunction($scope, $timeout, $anchorScroll, $location, $stateParams, SessionService, Assessment, Participant, Rubric) {
```
bithavoc: I'm trying to learn some angular here, is there a reason why we renamed this to **CtrlFunction**?

Also, is there a test for controller somewhere I'm not seeing?

spec/javascripts/controllers/AssessmentDashboardSidebarCtrlSpec.js
```diff
@@ -1,98 +1,99 @@
...
+
+      it('returns true if status is consensus', function() {
+        $scope.assessment.status = 'consensus';
+        expect($scope.consensusStarted()).toEqual(true);
+      });
     });
   });
-
-});
+})();
```
bithavoc: I noticed we performed formatting to this file but there is no test case for the new function `addLearningQuestion`, isn't this something we'd like to add to our test coverage?

app/assets/javascripts/client/controllers/learning_questions/LearningQuestionListDisplayCtrl.js
```diff
@@ -0,0 +1,61 @@
...
+          .get({assessment_id: vm.extractId()})
+          .$promise
+          .then(function (result) {
+            vm.learningQuestions = result.learning_questions;
+          });
+    };
+
+    vm.deleteLearningQuestion = function (model) {
+      var willDelete = $window.confirm('Are you sure you wish to delete this learning question?');
+      if (willDelete) {
```
bithavoc: You're performing a decision here that affects the rest of the scope, for such cases I prefer stopping the function execution ASAP.
```js
if (!willDelete) { return; }
```
This way if there is anything else we'd like to add in the future we'll add it to the end of the function instead the `if` branch, looks cleaner to me.

app/assets/javascripts/client/controllers/learning_questions/LearningQuestionListDisplayCtrl.js
```diff
@@ -0,0 +1,61 @@
...
+            .delete({assessment_id: $stateParams.id, id: model.id})
+            .$promise
+            .then(function () {
+              $scope.$emit('learning-question-change');
+            });
+      }
+    };
+
+    vm.updateLearningQuestion = function (model) {
+      if (model.editable) {
```
bithavoc: same here

app/controllers/v1/learning_questions_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    if learning_question.user == current_user
+      learning_question.destroy
+      render nothing: true, status: :no_content
+    else
+      @errors = 'You may not delete a learning question that does not belong to you.'
+      render 'v1/shared/errors', status: :bad_request
+    end
+  end
+
+  def exists
```
bithavoc: Is there any reasoning behind creating this operation instead of just using `#show`? You're basically doing the same.

bithavoc:
Tried to add a blank question but there is no clear indication of what's going on, maybe we're missing a validation there. Another option could be Disabling the button until the user types the question.

screen shot 2016-02-25 at 10 20 45 am

bithavoc:
Also noticed this while editing the question

bithavoc:
Trying to add a considerable large question doesn't give me any visual feedback that my text is too large for the question, looks like the limit is `255` chars.

Tried with the following paragraph:
```
 Although initially thought to have originated as a manmade variant of an extant canid species (variously supposed as being the dhole,[3] golden jackal,[4] or gray wolf[5]), extensive genetic studies undertaken during the 2010s indicate that dogs diverged from an extinct wolf-like canid in Eurasia 40,000 years ago.[6] Being the oldest domesticated animal, their long association with people has allowed dogs to be uniquely attuned to human behavior,[7] as well as thrive on a starch-rich diet which would be inadequate for other canid species.[8]
```

Results:

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
MakotoTheKnight: Not my code; part of what I was doing with that controller was conforming to [John Papa's style guide.](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md) 

Looks like the only changes in this particular class are centered around more style specific changes and moving the controller out in to its own function.  If you like, I could revert this specific file.

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
MakotoTheKnight: Not my code - this particular code was touched by code formatting changes.

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
bithavoc: gotcha, any chance we can remove it? is the view or the controller itself using it somehow?

spec/javascripts/controllers/AssessmentDashboardSidebarCtrlSpec.js
MakotoTheKnight: We could add some tests to ensure that it's using the right template, but that would be all we'd be testing for that particular modal.

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
MakotoTheKnight: The function actually is the controller, but since it's not my code, I will be happy to revert it to reduce any confusion.

app/assets/javascripts/client/controllers/assessments/AssessmentAssignCtrl.js
MakotoTheKnight: I named it that to get it away from repetition in names, but conventions that I've seen have had the named function be the same name as the controller.

app/assets/javascripts/client/controllers/learning_questions/LearningQuestionListDisplayCtrl.js
MakotoTheKnight: I'm personally not a fan of early returns, since it makes the code less appealing in my mind.  Further, the likelihood of us adding any additional functionality to that specific block is very slim; all we're doing is checking if the user agrees to delete the learning question, and if they don't, there's no further action required.

app/controllers/v1/learning_questions_controller.rb
MakotoTheKnight: It's not quite the same as a `#show` action.  We're checking with this particular method to see if the user has created _any_ learning questions for a given assessment.  Normally I've seen the `#show` action include the id used for the entity we're accessing; since we don't have a specific ID for a learning question at that point in time, we can't really include it.

app/assets/javascripts/client/controllers/learning_questions/LearningQuestionListDisplayCtrl.js
MakotoTheKnight: My comments earlier on early returns apply, but this is also a sanity guard.  We don't show the edit button or (if through DOM manipulation they managed to invoke it) allow the user to edit the content of the question, either.

This did highlight a case I didn't catch with the delete above; we also need to be sure that the entity is editable before we even show the confirm dialog.

MakotoTheKnight:
@bithavoc:  Go ahead and give this another look.

MakotoTheKnight:
The route `/v1/assessments/:id/permissions/all_users` still brings back the user with more than one possible role on it.

bithavoc:
LGTM

PR #101, opened 2016-02-27
Closes #81: inventory models

### LINKED ISSUE #81: Product Inventory:  Inventory Start
Rudimentary models:
- Inventory
 - name
 - deadline
 - district (associative)
 - has_many :product_entries
 - has_many :data_entries

- ProductEntry
(see attached document)

- DataEntry
(see attached document)

https://docs.google.com/document/d/1IWKZAtGon0QuZAXbswjOpc2fWX3xZ9MDmG0qHbivHS4/edit#

MakotoTheKnight:
This PR is good to go. Notes:

 - This creates a lot of models for us to operate with, and we take their values from the technical document.
 - There is a sizable amount of delegation with this; the intention is to not create the individual entities by themselves but rather have the parent Inventory class handle that work.  Preferably, the future controller would pass the attributes down that were needed to the inventory and let its children do the heavy lifting.

Let me know if there are questions/concerns.

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -16,6 +16,7 @@
 #  mandrill_id     :string(255)
 #  mandrill_html   :text
 #  report_takeaway :text
+#  share_token     :string
 #
 
 #  name            :string(255)
--- a/app/models/data_access_question.rb
+++ b/app/models/data_access_question.rb
@@ -10,6 +10,7 @@
 #  notes                :string
 #  created_at           :datetime
 #  updated_at           :datetime
+#  data_entry_id        :integer
 #
 
 class DataAccessQuestion < ActiveRecord::Base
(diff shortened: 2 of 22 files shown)


bithavoc:
I guess we can work with these models, my only concern with the enums we intent to use here is that they're hardcoded(require a developer to be enhanced/modified) and the admins won't be able to manage them dynamically via admin(we do have a rails_admin at `/admin`).

Do we want to make a decision on either the likelihood these lists would change or do we want to be consistent with the current features of the app and allow these lists to be modified via rails admin?

I know using enums is easier development-wise, but that's not necessarily what the client is expecting.

Also, unless I'm missing something from the Inventory Design document, it doesn't seem like we'll be making business decisions based on the user selection of items of these lists, this is all data entry and to me doesn't justify hardcoding them.

@candyj0311, thoughts?

MakotoTheKnight:
Closing this PR out since its merge target is incorrect.

PR #102, opened 2016-02-29
Closes #81: inventory models

### LINKED ISSUE #81: Product Inventory:  Inventory Start

MakotoTheKnight:
This PR is good to go. Notes:

 - This creates a lot of models for us to operate with, and we take their values from the technical document.
 - There is a sizable amount of delegation with this; the intention is to not create the individual entities by themselves but rather have the parent Inventory class handle that work. Preferably, the future controller would pass the attributes down that were needed to the inventory and let its children do the heavy lifting.

There are some concerns around the enums, their mutability, and Rails Admin's involvement, but we will wait until we have some information to go forward with any aspect to do with that.

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -16,6 +16,7 @@
 #  mandrill_id     :string(255)
 #  mandrill_html   :text
 #  report_takeaway :text
+#  share_token     :string
 #
 
 #  name            :string(255)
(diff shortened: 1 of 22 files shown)


MakotoTheKnight:
Concerns were covered [here](https://github.com/MobilityLabs/pdredesign-server/pull/101#issuecomment-190238349).

PR #103, opened 2016-02-29
Report controller spec cleanup

bithavoc:


spec/controllers/v1/report_controller_spec.rb
```diff
@@ -3,109 +3,191 @@
...
+    let(:assessment) { FactoryGirl.create(:assessment, :with_participants) }
+
+    context 'without user' do
+      before(:each) do
+        sign_out :user
+        get :show, assessment_id: 1, format: :json
+      end
+
+      it do
+        sign_out :user
```
MakotoTheKnight: We shouldn't have to sign out twice here.

spec/controllers/v1/report_controller_spec.rb
```diff
@@ -3,109 +3,191 @@
...
+        it 'returns participants as array as the participants of the assessment' do
+          expect(json["participants"].count).to eq(assessment.participants.count)
+        end
+
+        it 'get the participant count' do
+          expect(json["participant_count"]).not_to be_nil
+        end
+      end
+
+      context 'on an existing consensus' do
```
MakotoTheKnight: This should probably be a non-existing consensus.

spec/controllers/v1/report_controller_spec.rb
```diff
@@ -3,109 +3,191 @@
...
-  before :each do
-    request.env["HTTP_ACCEPT"] = 'application/json'
-  end
+  describe '#show' do
+    let(:assessment) { FactoryGirl.create(:assessment, :with_participants) }
+
+    context 'without user' do
+      before(:each) do
+        sign_out :user
+        get :show, assessment_id: 1, format: :json
```
MakotoTheKnight: Even though we're going to fail this case regardless of what ID we pass through, it's better in terms of behavior and expressiveness to use the object that we intended here.  This way, if we see a corresponding successful test that uses the same assessment object, we can build more confidence in that this is testing the right thing.

MakotoTheKnight:
The few comments I left aren't showstopping, and I'm fine with these tests as they are.  :shipit:

PR #107, opened 2016-03-01
Closes #100: Restrict learning questions

### LINKED ISSUE #100: User Cannot Add Learning Question After Consensus Meeting Date
Users cannot add Learning Questions once the consensus meeting date has passed. 

Remove button in sidebar to add Learning Questions once date has passed.

MakotoTheKnight:

 - We don't allow "completed" states to see or add to the list of existing questions.
 - We show a modal if a user has not yet created a learning question, and continue to show it until one is created (this was the original requirement in #79).
 - This fixes an error with the delete - we did not change around the ID for deletion, which meant that the request to delete an entity would fail.


app/assets/javascripts/client/controllers/responses/ResponseCtrl.js
```diff
@@ -1,14 +1,38 @@
...
+    $scope.assessment = Assessment.get({id: $stateParams.assessment_id});
+
+
+    $scope.$on('responses-loaded', function() {
+      $timeout(function() {
+        LearningQuestion.exists({assessment_id: $stateParams.assessment_id})
+            .$promise
+            .then(function() {
+              // No-op - implies that a learning question exists for this user on this assessment
+            }, function() {
```
bithavoc: shouldn't we check the error of this call to know if it's exactly 404 what we got here?

app/assets/javascripts/client/directives/learning_questions/learningQuestionModal.js
```diff
@@ -0,0 +1,24 @@
...
+  function learningQuestionModal() {
+    return {
+      restrict: 'E',
+      scope: {
+        reminder: '@'
+      },
+      templateUrl: 'client/views/modals/learning_questions/add_learning_question.html',
+      link: function(scope) {
+        // The modal exists on the parent scope, so we have to reach up to it  to bind any closing actions on the modal.
+        var parentScope = scope.$parent;
```
bithavoc: How safe and clean is this compared to emitting an event and capture it in the parent scope to dismiss the modal?

app/assets/javascripts/client/controllers/responses/ResponseCtrl.js
MakotoTheKnight: I originally intended that endpoint to be a HEAD request, but our current version of Rails and HEAD requests don't seem to mix all that well.  In the context of the head, all I'm checking for is if it exists or not; if it doesn't then we should be opening the modal.

I agree here though that we should probably ensure that we're not getting a 500 back from the promise which would cause the modal to open.  A quick smoke test of this would be to attempt to load the page, and while it loads, turn off the server to see 500s come back.  The modal ideally shouldn't open if we get any 500s back.

bithavoc: I'm not following, why 500 specifically? aren't there any  other status code we care about? where are we checking this?

app/assets/javascripts/client/directives/learning_questions/learningQuestionModal.js
MakotoTheKnight: Emitting the event would be a fair bit tougher to track down and debug.  Here, we're making it explicit that the parent contains the modal, and that we require the parent scope to do so.

app/assets/javascripts/client/controllers/responses/ResponseCtrl.js
MakotoTheKnight: In the context of a HEAD request, we would either get back a 204 (No Content) or a 404 (Not Found), which are the only two states we should generally care about with this request.  That is to say, when we make the request, all we're checking to see is if we get a response back in the affirmative for the condition being true.

If we get back a 5xx error (most of the errors I've seen from Rails recently have been 500s, which is where my mind was at the time), then something on the server side has erred.  We shouldn't display the modal in the event of this, since we can't determine if the entity does or does not exist due to the error.

app/assets/javascripts/client/directives/learning_questions/learningQuestionModal.js
bithavoc: Under what circumstances would this break if the parent doesn't contain the modal? I'm afraid this explicitness would limit the way we use this specific directive, it's making a lot of assumptions about the parent scope, I'm not sure if this is natural to Angular.js but in other lands isn't this called feature envy?

MakotoTheKnight: That's a fair point.  I can change it around to use the event.

app/assets/javascripts/client/controllers/responses/ResponseCtrl.js
bithavoc: Makes sense :+1:

app/assets/javascripts/client/directives/learning_questions/learningQuestionModal.js
MakotoTheKnight: I've went back to this code and looked a bit closer at what it would take to convert this over to using events.

There are about four different spots that this modal is being created in which emitting the event would require us to place the same code in four different spots:

```
$scope.$on('close-learning-question-modal', function() {
  $scope.modal.dismiss('cancel');
});
```

The event code isn't complex but the number of places that this has to be placed in is higher than I'd like it to be.

bithavoc: I don't think this is a problem at all, actually, in observer patterns is pretty normal to have multiple listeners, this is better than hardcoding the name of the variable in the parent scope, now `AssessmentDashboardSidebarCtrl`, `ResponseCtrl` and `SidebarResponseCardCtrl` would have a chance to freely rename it's modal variable at will without having to affect `learningQuestionModal`. you just gave `learningQuestionModal` one less reason to change, that's awesome.

bithavoc:
LGTM

PR #110, opened 2016-03-03
Adjusting the inventory models schema

MakotoTheKnight:

 - Removes unnecessary IDs in parent Inventory model
 - Add missing belongs_to relationships
 - Add spec tests to cover said relationships
 - Change Response.create! in util_helper to Response.find_or_create_by!

### DIFF
--- a/app/models/inventory.rb
+++ b/app/models/inventory.rb
@@ -15,8 +15,4 @@ class Inventory < ActiveRecord::Base
 
   accepts_nested_attributes_for :product_entries
   accepts_nested_attributes_for :data_entries
-
-  default_scope {
-    includes(:product_entries, :data_entries)
-  }
 end
--- a/app/models/inventory.rb
+++ b/app/models/inventory.rb
@@ -2,23 +2,21 @@
 #
 # Table name: inventories
 #
-#  id               :integer          not null, primary key
-#  name             :string           not null
-#  deadline         :datetime         not null
-#  district_id      :integer          not null
-#  product_entry_id :integer
-#  data_entry_id    :integer
+#  id          :integer          not null, primary key
+#  name        :string           not null
+#  deadline    :datetime         not null
+#  district_id :integer          not null
 #
 
 class Inventory < ActiveRecord::Base
-  has_one :product_entry
-  has_one :data_entry
+  has_many :product_entries
+  has_many :data_entries
   belongs_to :district
 
-  accepts_nested_attributes_for :product_entry
-  accepts_nested_attributes_for :data_entry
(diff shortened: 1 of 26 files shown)


app/models/inventory.rb
```diff
@@ -2,23 +2,21 @@
...
+#  name        :string           not null
+#  deadline    :datetime         not null
+#  district_id :integer          not null
 #
 
 class Inventory < ActiveRecord::Base
-  has_one :product_entry
-  has_one :data_entry
+  has_many :product_entries
+  has_many :data_entries
```
bithavoc: This makes a lot more sense now... :tada:

app/models/inventory.rb
```diff
@@ -2,23 +2,21 @@
...
   belongs_to :district
 
-  accepts_nested_attributes_for :product_entry
-  accepts_nested_attributes_for :data_entry
+  accepts_nested_attributes_for :product_entries
+  accepts_nested_attributes_for :data_entries
 
   default_scope {
-    includes(:product_entry, :data_entry)
+    includes(:product_entries, :data_entries)
```
bithavoc: mmm I'm not sure if we'd all all these entries to the default scope, if we're showing all the inventories in a list for the user to pick we don't necessarily need the product entries for that, this was fine before when we had 1-1 relationship but not that is 1-to-many it feels a little odd to me.

MakotoTheKnight: That's a fair point.  We can go ahead and remove them.

bithavoc:
Does it still makes sense to have `ProductEntry` associations as `has_one`? if we do, why is that'

MakotoTheKnight:
From the designs, the form for a product entry (and data entry, for that matter) is one massive sheet of information.  I didn't see any place that would allow one to create multiple general inventory questions or multiple technical questions.

bithavoc:
I see, how would a factory looks like with `has_one`, how could the factory for `product_entry` would look like for example? I want to see how to prevent a loop from happening.

MakotoTheKnight:
I think I see where you're going with that.  The intent is surely to avoid a scenario in which we have a cyclic dependency.  Let me take a closer look at these has_one/belongs_to relationships on the children entities.

bithavoc:
remember what we talked about making sure the model can be actually saved? I kinda did something similar with `Inventory` just to make sure they work, I was wondering if we could do the same thing with all the models involved, it shouldn't take long.

MakotoTheKnight:
@bithavoc: I moved the creation of the enum value out of the trait since, by default, we should just be getting a value back.

bithavoc:
LGTM

Issue #83, opened 2016-02-15
Product Inventory:  Add District Users to Inventory


MakotoTheKnight:
Borrow from Steps 2 and 3 from the existing assessments flow and allow the facilitator to manage the participants and send an email out to others.

Attached image for reference:
 
![pasted image 0](https://cloud.githubusercontent.com/assets/1821302/13061138/2d2aa9c8-d3f3-11e5-8a63-763956047be7.png)

A facilitator must also be able to send reminders to all participants in case no one has responded.  This functionality is similar to what we already have in place for Assessments.

candyj0311:
Wise words from Jason: make sure we're not copy-pasting wholesale the Readiness Assessment for this piece.

candyj0311:
![image](https://cloud.githubusercontent.com/assets/10750162/13409181/c1390274-deff-11e5-8473-e94899b0601a.png)

candyj0311:
![image](https://cloud.githubusercontent.com/assets/10750162/13409326/7b402666-df00-11e5-9c5c-8d9916fe6d55.png)

bithavoc:
This issue will only focus on:

* When there district users that are not yet members of the inventory, it will show a link that triggers a modal listing all the district users that can be added to the inventory
* Only users from the district that are not network partners and haven't been added to the inventory can be added
* Once all intended users have been added from the district, modal can be close via **Done** or **X** button (upper right side of the dialog)
* The button *Add Users* currently look as a clickable label until we apply the style as part of https://github.com/MobilityLabs/pdredesign-server/issues/132

PR #112, opened 2016-03-04
Server side support for Inventory Invitations

bithavoc:
Server side support for #83


spec/controllers/v1/inventory_invitations_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+        it 'returns error message' do
+          expect(json["errors"]["email"]).to include("User has already been invited")
+        end
+      end
+
+      context 'inviting user for first time' do
+
+        let(:created_invitation) { InventoryInvitation.where(email: email).first }
+
+        shared_examples 'successful_invitation' do
```
MakotoTheKnight: Could you move this to a separate file?

lib/inventories/permission.rb
```diff
@@ -0,0 +1,107 @@
...
+      @inventory = inventory
+      @user = user
+    end
+
+    def possible_roles_permissions
+      PERMISSIONS-[get_level(user)]
+    end
+
+    def requested
+      AccessRequest.where(assessment_id: assessment.id)
```
MakotoTheKnight: May want to remove references to assessment.  We will likely need to change it to inventory, but let's leave them out for now.

lib/inventories/permission.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    def get_access_request(user)
+      AccessRequest.find_by(assessment_id: assessment.id, user_id: user.id)
```
MakotoTheKnight: This model - `AccessRequest` - may be a good candidate to become polymorphic to increase its reusability.

app/controllers/v1/inventory_invitations_controller.rb
```diff
@@ -0,0 +1,41 @@
+class V1::InventoryInvitationsController < ApplicationController
+  before_action :authenticate_user!
+  authorize_actions_for :inventory
+
+  authority_actions create: 'update'
+  def create
+    create_params = invitation_params
+    create_params[:inventory_id] = inventory_id
+
+    send_invite = create_params.delete(:send_invite)
```
MakotoTheKnight: This parameter surprised me, but I then realized that we were doing something similar [in the `ParticipantsController`](https://github.com/MobilityLabs/pdredesign-server/blob/master/app/controllers/v1/participants_controller.rb#L7).  What ultimately tipped me off was that we were passing in a boolean to send emails, although in practice, the check later only checks for a truthy value.

The likely purpose of this is to allow us to save without us automatically triggering emails, but if we don't provide that parameter, we introduce the state of us saying we've sent the email, but that hasn't actually happened.  This makes resending the email with the same parameters not currently possible.

Do we need this functionality right now?  I could see us adding it in later, but I'm not so sure we need it at this point in time.

app/models/inventory.rb
```diff
@@ -6,13 +6,38 @@
 #  name        :string           not null
 #  deadline    :datetime         not null
 #  district_id :integer          not null
+#  owner_id    :integer
 #
 
 class Inventory < ActiveRecord::Base
+  include Authority::Abilities
```
MakotoTheKnight: I skimmed through the documentation on Authority and I feel pretty good about it, but it's not immediately apparent that we directly call any of its methods for an `Inventory` model.  Is it the case that this is done transparently at some layer?

app/models/inventory.rb
```diff
@@ -6,13 +6,38 @@
...
   belongs_to :district
+  belongs_to :owner, class_name: 'User'
+  self.authorizer_name = 'InventoryAuthorizer'
 
   accepts_nested_attributes_for :product_entries
   accepts_nested_attributes_for :data_entries
+
+  has_many :members, class_name:'InventoryMember'
+  has_many :participants, -> { where(role: 'participant') }, class_name:'InventoryMember'
+  has_many :facilitators, -> { where(role: 'facilitator') }, class_name:'InventoryMember'
```
MakotoTheKnight: I think we had a conversation about this before - the user's role is apparently being mutated globally and not per assessment (that is to say, a user cannot hold two distinct roles in two distinct assessments).  This seems like a clean approach to get us away from that.

app/models/user.rb
```diff
@@ -72,5 +72,9 @@ def email=(value)
   def name
     "#{first_name} #{last_name}"
   end
- 
+
+  def ensure_district(district:)
```
MakotoTheKnight: Couple of points here:

 - This method is mutating the state of the User instance, so it likely should end with a `!`.
 - The name is a bit unusual to me; it's not exactly ensuring that there's a district with that value there, it's adding one if if it doesn't exist.  Perhaps it should be `conditionally_add_district!` to better convey its intent?

db/schema.rb
```diff
@@ -432,7 +467,7 @@
   add_index "scores", ["response_id", "question_id"], name: "index_scores_on_response_id_and_question_id", unique: true, using: :btree
 
   create_table "sessions", force: :cascade do |t|
-    t.string   "session_id", null: false
+    t.string   "session_id", limit: 255, null: false
```
MakotoTheKnight: How'd this change?

lib/inventories/member_from_invite.rb
```diff
@@ -0,0 +1,73 @@
...
+      user.save
+      return user
+    end
+
+    def ensure_user_district(user)
+      user.ensure_district(district: inventory.district)
+    end
+
+    def create_user
+      User.create!(first_name:   invite.first_name,
```
MakotoTheKnight: This seems to be a problem.  If the user is coming from outside of the system, and their account is created through this invite, then they can not register a new account.

What was the use case if we couldn't find a user in the system?

lib/inventories/member_from_invite.rb
```diff
@@ -0,0 +1,73 @@
...
+                   password:     generate_password,
+                   team_role:    invite.team_role)
+    end
+
+    def user_found
+      @user_found = User.find_by(email: invite.email)
+    end
+
+    def generate_password
+      SecureRandom.hex[0..9]
```
MakotoTheKnight: It's not immediately apparent what this code is doing, but it seems to be the rough equivalent of `SecureRandom.hex(5)`.

spec/models/inventory_spec.rb
```diff
@@ -23,4 +26,43 @@
```
MakotoTheKnight: We missed the tests where we expect to have many facilitators and many participants.

MakotoTheKnight:
When attempting to create a PATCH request for an ID that no longer exists, we get back the whole error message from Rails as opposed to a JSON error.

The culprit is here:

    inventory.access_requests.find(access_request_id)

I also didn't expect the entity to have been removed from the database at this point in time; I had sent a previous POST to create the request, and a PATCH request to approve it beforehand.

spec/controllers/v1/inventory_invitations_controller_spec.rb
bithavoc: where'd be an  ideal place to put these, inside `support`?
Should I make them load automatically or `require` them explicitely?

app/models/user.rb
bithavoc: * last time I check a method that mutates does not necessarily have to be a bang method, I thought it was intended for `danger` methods and can't really see what's makes this method dangerous.
* I thought `ensure` in the prefix would be explicit enough and it's in fact adding the district only when the user doesn't have the district added to it.
* the name you suggest `conditionally_add_district` doesn't really tell me what's the condition is and it seems to be leaking part of the implementation in the name of the method.

db/schema.rb
bithavoc: I have no idea what happened here, I'd try to restore the schema from staging and reapply migrations.

lib/inventories/member_from_invite.rb
bithavoc: yeah, not my code, I borrowed this code from assessments and didn't noticed. Will fix

spec/models/inventory_spec.rb
bithavoc: :+1:

bithavoc:
@MakotoTheKnight fixed the `not_found` error. about destroying the access request, it makes sense it's destroyed once it's being `granted` or `denied` so the user is able to request access again later on. this is how assessments works now.

db/schema.rb
bithavoc: This change is already present in `staging` branch.

lib/inventories/member_from_invite.rb
bithavoc: this method is being called on `create_or_update_user`, it won't be called if the user already exists.

The user needs to be created prior sending the invite, we got our answer at the meeting yesterday.

MakotoTheKnight:
I think I was surprised by its removal due to there being no indication in the response that the request had been fulfilled to my satisfaction (i.e. no helpful response indicating what had transpired).

For the most part, though, I'm okay with what I see here. Let's go ahead and merge it in.

MakotoTheKnight:
:shipit:

PR #113, opened 2016-03-07
V2 style initial

MakotoTheKnight:
This PR is the same as #92, with the key difference being that this PR is pointing to staging instead of master.


MakotoTheKnight:
I'm noticing some issues with the footer:

image

This should likely be captured in a separate task and worked on there.

PR #117, opened 2016-03-08
Hotfix: update due date

MakotoTheKnight:
This PR is good to go. Notes:

 - This specifically addresses an issue in which we have two nodes on the DOM with the id 'due-date'.  This is due to the application HTML rendering elements when it shouldn't be (it was using ng-hide to visually hide them from the display instead).


MakotoTheKnight:
@bithavoc:  When you get some free time, could you give this a review?

bithavoc:
LGTM

PR #129, opened 2016-03-15
Closes #111: Creating a new Product Inventory

### LINKED ISSUE #111: Product Inventory:  Creating a Product Inventory
Create the main landing page where a user comes in to start the product inventory.

For now, we will be borrowing the same flow as creating an assessment.

MakotoTheKnight:

 - This largely borrows from the assessment creation dialog, but there are some distinctions with it in that we always expect a district to exist for the user that is creating it.  If we need to adjust this logic, then we can opt to hide the creation for users who would not have a district attached.
 - There are a lot of reorganization bits in this commit; namely, we're moving around the Angular code away from a grouping by type to a grouping by role.
 - Some of the Angular controllers and directives were broken up, which preserved their behavior but also shifted around their respective code density.
 -  Certain directives that were touched were done so to introduce IIFEs and remove unused injections.


app/assets/javascripts/client/common_list/EntryCtrl.js
```diff
@@ -0,0 +1,116 @@
+(function() {
+  'use strict';
+
+  angular.module('PDRClient')
+      .controller('EntryCtrl', EntryCtrl);
```
bithavoc: Made a quick look for `EntryCtrl` in this PR but couldn't really tell what it does, the name is not very specific to an implementation so I can assume it's an `abstraction`, however, it does has members like `consensusReportLink` and `assessmentLink` that seem to be very specific to a *concretion* of assessments. maybe this is breaking the DIP principle (?)

app/assets/javascripts/client/common_list/EntryCtrl.js
```diff
@@ -0,0 +1,116 @@
...
+        case 'response':
+          return '/assessments/' + entity.id + '/responses/' + entity.responses[0].id;
+        case 'consensus':
+          return '/assessments/' + entity.id + '/consensus/' + entity.consensus.id;
+        case 'none':
+          return false;
+      }
+    };
+
+    vm.orderLinks = function(items) {
```
bithavoc: is it the case that both `orderLinks` and `responseLink` are trying to supply computed values that are specific to missing pieces in `entity`? at least I can see `items` looking like the result of a previous operation, maybe these computed values should be computed before `items` is returned wherever it came from.

app/assets/javascripts/client/home/HomeCtrl.js
```diff
@@ -0,0 +1,83 @@
...
+      return '<p class="no-link">This item is currently under development. Please stay tuned.</p>';
+    };
+
+    $scope.popoverContent = function(tool) {
+      if (!tool.description)
+        return $scope.emptyTool();
+
+      var output = "<div><div class='row'><div class='col-md-12'><p class='greeting'>" +
+          tool.title + "</p>" + '<p class="description">' + tool.description + '</p>' + '<p>';
+      if (!tool.url)
```
bithavoc: I wonder if a controller be spitting HTML, maybe we can accomplish the same goal with `ng-hide` in a template.

app/assets/javascripts/client/home/StartAssessmentCtrl.js
```diff
@@ -0,0 +1,97 @@
...
+    var vm = this;
+
+    vm.user = SessionService.getCurrentUser();
+    vm.district = vm.user.districts[0];
+
+    vm.alerts = [];
+    vm.assessment = {};
+
+    vm.hideModal = function() {
+      $('#startAssessment').modal('hide');
```
bithavoc: I wonder if we should be using `$element` instead, otherwise the DOM element won't be found while running the test scenarios. I also don't see any tests for this method, do we need it?

app/assets/javascripts/client/common_list/EntryCtrl.js
MakotoTheKnight: This is existing code.  I'm no fan of it either, but this is how it originally came.  I preserved it here to not cause any regressions.

app/assets/javascripts/client/home/HomeCtrl.js
MakotoTheKnight: Two things:

 - If we want this to show conditionally we'd really rather use ng-if instead.
 - This is also existing code, ported over to preserve its original behavior.

app/assets/javascripts/client/home/StartAssessmentCtrl.js
MakotoTheKnight: This is existing code but the catch here is that this is using the Bootstrap CSS approach to modals instead of the `$modal` service, making it a lot harder to test.  A better rewrite of this would involve using the actual `$modal` service, which would make this more straightforward to test.

bithavoc:
Is it going to be a problem that is not really clear what the second select/combobox is really about?

MakotoTheKnight:
Fair comment - that's how assessments have always looked (there's no default option), but I defer to @rickymetz on this one.

bithavoc:
I created a new third tech audit, hit `Enter` but even when the dialog disappears (I guess this is expected), the list doesn't refresh until I hit Refresh in my browser. My first impression was that some kind of error occurred and I may feel tempted to proceed to create the tech audit all over again.



is this a problem?

bithavoc:
I just created a new tech audit with an specific date, however, the tech audit list is reporting a different deadline date.

MakotoTheKnight:
@bithavoc: 
I'll double-check that.  I did notice a spot where my mind went on autopilot with DD/MM/YYYY format.  What input did you use in the form?

bithavoc:
@MakotoTheKnight I didn't use any input except for the datepicker (I'm that lazy :trollface:  )

app/assets/javascripts/client/home/StartAssessmentCtrl.js
```diff
@@ -0,0 +1,97 @@
...
+      return _.isEmpty(vm.user.district_ids);
+    };
+
+    vm.redirectToAssessment = function(assessment) {
+      vm.hideModal();
+      $location.path('assessments/' + assessment.id + '/assign');
+    };
+
+    vm.create = function(assessment) {
+      assessment.due_date = moment($("#due-date").val(), 'MM-DD-YYYY').toISOString();
```
bithavoc: this is new to me, how are we doing other parsing around the project? do we want this formatting style hardcoded here? what do you think about having them in a `dateService` or somewhere centralized?

app/assets/javascripts/client/home/StartAssessmentCtrl.js
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    vm.openAssessmentModal = function() {
+      vm.assessmentModal = $modal.open({
+        templateUrl: 'client/home/assessment_modal.html',
+        scope: $scope
+      });
+    };
+
+    SessionService.syncUser();
```
bithavoc: I see this in a lot of places of the project, what does it do and why does it apply for this case?

app/assets/javascripts/client/home/StartAssessmentCtrl.js
MakotoTheKnight: I'd love to unify this at some point as well since we should only have one way to do this.  But it's likely not in the scope of this task.  This looks like a hold over from the previous controller that existed.

app/assets/javascripts/client/home/StartAssessmentCtrl.js
MakotoTheKnight: I can't say I'm sure.  This is legacy code that was likely shuffled around.  I didn't touch it since I didn't wish to break its behavior.

app/assets/javascripts/client/home/StartAssessmentCtrl.js
bithavoc: I understand that, but honoring the same spirit of previous code-reviews in this same project, it's likely we'd like to fix this now since you previously suggested fixes on code that was being borrowed from other parts of the project that don't always feel entirely right. I'm not suggesting fixing other parts of the projects with this new service, I'm just suggesting creating a service and use it here now, this way another developer who wants to borrow your code doesn't have to fix it/duplicate it.

bithavoc:
It looks like we're structuring this feature in a different way than other parts of the project, at what point did we started switching from our current scheme `controllers/<feature>.js` to `<feature>/controller.js`?

app/controllers/v1/inventories_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    unless inventory_params[:deadline].blank?
+      begin
+        @inventory.deadline = DateTime.strptime(inventory_params[:deadline], '%m/%d/%Y')
+      rescue
+        @inventory.errors.add(:deadline, 'must be in DD/MM/YYYY format')
+        return render_error
+      end
+    end
+    @inventory.district = District.find(inventory_params[:district][:id])
+    @inventory.user = current_user
```
bithavoc: you can also do `@inventory = current_user.inventories.new(inventory_params.extract!(:name))`

spec/controllers/v1/inventories_controller_spec.rb
```diff
@@ -0,0 +1,267 @@
...
+
+  describe 'GET #index' do
+    context 'when not authenticated' do
+
+      before(:each) do
+        get :index, format: :json
+      end
+
+      it {
+        expect(response.status).to eq 401
```
bithavoc: I personally find `expect(response).to have_http_status :not_found` to be more clear and **cool** :smile: but this is doing the trick too.

MakotoTheKnight:
I've updated this with the latest fix to the date format @bithavoc; mind taking another look?

bithavoc:
This is an interesting scenario, I just created my account and don't have a district set in my profile yet, trying to provision a new Tech Audit wouldn't allow me to select anything and there is no clear indication of what to do in such cases.

MakotoTheKnight:
I would imagine that it should behave just like an assessment would in that it forces you to create a district before it continues.  That's something that can be addressed.

bithavoc:
FYI, for assessments we show instructions for the user.

app/assets/javascripts/client/common_list/EntryCtrl.js
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    vm.responseLink = function(entity) {
+      switch (entity.response_link) {
+        case 'new_response':
+          return '/assessments/' + entity.id + '/responses';
+        case 'response':
+          return '/assessments/' + entity.id + '/responses/' + entity.responses[0].id;
+        case 'consensus':
+          return '/assessments/' + entity.id + '/consensus/' + entity.consensus.id;
```
bithavoc: how many of this routes items apply for tech audit?

MakotoTheKnight: None of it for now.  That was a straight-up port from the existing code.  In the future I anticipate that it will have a lot to do with it, but for now it doesn't.

bithavoc: so, we're basically staging here that **Tech Audit** and **Assessment** have nothing in common, however, this file will serve both purposes/concepts?

bithavoc:
I can’t really say I'm ok with `entry` and `common_list`, I don’t think the name is correct because it looks like a generic/abstract file/directive for entries in a list but it has a lot of concretions in it related to assessment (and apparently Tech Audit in the future).

@MakotoTheKnight seems like you're ok with it, feel free to merge it in.

bithavoc:
Functionality-wise this PR seems like it's doing the trick @MakotoTheKnight

PR #136, opened 2016-03-18
Automatically add inventory owner as facilitator

bithavoc:
The owner of the inventory should automatically become a member facilitator of the inventory upon creation, this cover most of the scenarios involving all facilitators of the inventory for notifications and  authorization.

We want to have this common behavior in place by default to avoid [spreading it all around the project](https://github.com/MobilityLabs/pdredesign-server/blob/staging/app/workers/access_request_notification_worker.rb#L17).

### DIFF
--- a/app/models/inventory.rb
+++ b/app/models/inventory.rb
@@ -33,6 +33,8 @@ class Inventory < ActiveRecord::Base
   has_many :participants, -> { where(role: 'participant') }, class_name:'InventoryMember'
   has_many :facilitators, -> { where(role: 'facilitator') }, class_name:'InventoryMember'
 
+  after_create :add_facilitator_owner
+
   def facilitator?(user:)
     facilitators.where(user: user).exists?
   end
@@ -47,5 +49,11 @@ class Inventory < ActiveRecord::Base
 
   def member?(user:)
     self.members.where(user: user).exists?
+  end
+
+  private
+  def add_facilitator_owner
+    return unless owner
+    facilitators.create(user: owner)
   end
 end
--- a/spec/models/inventory_spec.rb
+++ b/spec/models/inventory_spec.rb
@@ -52,11 +52,13 @@ describe Inventory do
   end
 
   describe '#save' do
-    subject {
-      create(:inventory)
-    }
+    let(:owner) { FactoryGirl.create(:user) }
+    subject { FactoryGirl.create(:inventory, owner: owner) }
 
-    it { expect(subject.new_record?).to be false } 
+    it { expect(subject.new_record?).to be false }
+    it 'add owner as facilitator' do
+      expect(subject.facilitators.where(user: owner)).to exist
+    end
   end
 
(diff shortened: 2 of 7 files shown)


bithavoc:
@MakotoTheKnight this should be a quick one

bithavoc:
@MakotoTheKnight please take a look when you get some time

bithavoc:
@MakotoTheKnight rebased, can you take a look?

MakotoTheKnight:
I'll take a look at this just before lunch.

bithavoc:
@MakotoTheKnight I'm kinda needing this now, let me know if you have some time to review this today.

app/models/inventory.rb
```diff
@@ -48,4 +50,10 @@ def owner?(user:)
   def member?(user:)
     self.members.where(user: user).exists?
   end
+
+  private
+  def add_facilitator_owner
+    return unless owner
+    facilitators.create(user: owner)
```
MakotoTheKnight: I'm thinking this is fine, but two questions come up:

 - How do we guard against repeated entries of the same owner?  Anecdotally I can't add seem to add the same owner as a facilitator, but I don't see a test covering this scenario.

 - I noticed that the `invited_at` date is `nil` for a facilitator created in this way.  Ideally, would we want to say that they were invited at the same time this was created (or `Time.now`)?

spec/models/inventory_spec.rb
```diff
@@ -52,11 +52,13 @@
   end
 
   describe '#save' do
-    subject {
-      create(:inventory)
-    }
+    let(:owner) { FactoryGirl.create(:user) }
+    subject { FactoryGirl.create(:inventory, owner: owner) }
```
MakotoTheKnight: This seems to be testing the behavior of the factory more than the model.  It will likely shake out, but if the factory does fail, we'll get the failure not just here, but everywhere that uses it.

app/models/inventory.rb
bithavoc: members are allowed to have `invited_at` as nil, for example, when you're added as a district user.

We don't need to check if the owner already exists as facilitators since this method is only called with `after_create`.

spec/models/inventory_spec.rb
bithavoc: the factory should be calling `save`, if it calls `save` we'll get `after_create` calling `add_facilitator_owner` which is the default behavior for factories and non-factories.

I'm not sure what your point is.

MakotoTheKnight: If the factory fails to save the entity, then everywhere that we make use of this factory will also fail for the same reason.  It's more for us to be sure that we isolate failures, but for now it will suffice.

app/models/inventory.rb
MakotoTheKnight: :+1:

MakotoTheKnight:
Alright, things look alright here to me.  :shipit:

PR #145, opened 2016-03-22
Closes #131:  Split up Facilitation page into bite-sized components

### LINKED ISSUE #131: Product Inventory: Create Page to Product Inventory: Break out elements on assign.html into components
Generify the existing assessment facilitation page and use it for product inventories.

This should be the location where new product inventories land after they are created.

MakotoTheKnight:
This PR is ready to review.  Notes:

 - This moves the logic of Tool Creation, Invites, and Participant management into their own directives.
 - No behavior should have changed during this move.
 - This PR also contains a fix in which we accidentally clobbered some sharing functionality in the sidebar, which is now safely restored.

A lot of these methods are still specific to Assessment for now, but they should be in a good spot to migrate over to utilize any tool we wish.


MakotoTheKnight:
@bithavoc:  When you have some time, could you look into this PR?

bithavoc:
@MakotoTheKnight is this all new code?

MakotoTheKnight:
This code is the same as before, with the notable exception that each component of the steps - updating the name, inviting the participants, updating the email, etc - are all broken out in to directives, as opposed to them being driven by a single controller.

There were some other changes to incorporate style (use of IIFE and `$inject`), but there should be no functional difference between this code and the previous code.  This only makes it easier for us to reuse the components on another page.

bithavoc:
LGTM

Issue #144, opened 2016-03-21
Product Inventory:  Create Page


MakotoTheKnight:
Use the split-up modals from #131 to make this happen.

The create page should behave and look identical to the current create page for assessments.  The main changes to this page would be the resource being used; instead of using an Assessment, we would be using an Inventory.

 - [x] Add show route for Inventories
 - [x] Update view to remove duplicate buttons on both Inventory and Assessment
 - [x] Create inventory assign page
 - [x] Add support for inventories to CreateService
 - [x] Add inventory support to LearningQuestion
 - [ ] Add simple navigation for draft entries _only_
 - [x] Redirect inventory creation to this page

candyj0311:
Add details once 131 is reviewed.

MakotoTheKnight:
Wire in the functionality captured in pull #148 to this page.  The inventory should be using that functionality and behavior.

MakotoTheKnight:
Depends on #142, #143, and #155.

PR #148, opened 2016-03-23
Closes #83: inventory add users

### LINKED ISSUE #83: Product Inventory:  Add District Users to Inventory

bithavoc:
Depends on https://github.com/MobilityLabs/pdredesign-server/pull/136


MakotoTheKnight:
Seeing this when I go to add participants to an existing inventory, or to add participants to a new inventory.

image

app/assets/javascripts/client/inventories/add_inventory_users_link.html
```diff
@@ -0,0 +1,3 @@
+<div>
+  <div ng-click="link.open()">Add Users</div>
```
MakotoTheKnight: This should definitely be a hyperlink.

bithavoc:
@MakotoTheKnight we may need an empty placeholder when there are no other users in the district that you could add. but that just indicates there are no district users available to add into the inventory.

app/assets/javascripts/client/resources/InventoryInvitable.js
```diff
@@ -0,0 +1,8 @@
+PDRClient.factory('InventoryInvitable', ['$resource', 'UrlService', function($resource, UrlService) {
```
MakotoTheKnight: This resource isn't using the IIFE style, and it's relying on the `PDRClient` variable.

app/assets/javascripts/client/resources/InventoryParticipant.js
```diff
@@ -0,0 +1,7 @@
+PDRClient.factory('InventoryParticipant', ['$resource', 'UrlService', function($resource, UrlService) {
```
MakotoTheKnight: This resource isn't using the IIFE style, and it's relying on the PDRClient variable.

app/assets/javascripts/client/resources/InventoryInvitable.js
bithavoc: :+1:

app/assets/javascripts/client/resources/InventoryParticipant.js
bithavoc: :+1:

app/assets/javascripts/client/inventories/add_inventory_users_link.html
bithavoc: where can I see the style of the links to be used here? we're going to use there in the inventory details page, it should match. do you have it?

MakotoTheKnight:
@bithavoc:  What's the equivalent behavior for when there are no district partners on an assessment?

bithavoc:
@MakotoTheKnight shows an empty list :disappointed:

bithavoc:
@MakotoTheKnight I'll an empty placeholder with message similar to the one we show for districts, how does that sounds?

app/assets/javascripts/client/common_list/entry_view.html
```diff
@@ -58,6 +58,7 @@ <h3 class='name'>{{entity.name}}
         <span class="stat">{{entity.participant_count}}</span>
         <span class="category">Completed</span>
         <span class="stat">{{entity.completed_responses}}</span>
+        <add-inventory-users-link data-inventory-id="entity.id"></add-inventory-users-link>
```
MakotoTheKnight: We don't need the `data-` prefix on the parameter here.  Since we're replacing the custom HTML with the directive, there's no need to worry about using `data-`.

bithavoc: :+1:

app/assets/javascripts/client/inventories/add_inventory_users_modal.html
```diff
@@ -0,0 +1,12 @@
+<div class="request-access-modal">
+  <div class="modal-header">
+    <button ng-click="link.close()" type="button" class="close" aria-hidden="true">×</button>
+    <h2>Inventory Users</h2>
+  </div>
+  <add-inventory-users data-inventory-id="inventoryId"></add-inventory-users>
```
MakotoTheKnight: Same as before - no need to have the `data-` prefix.

bithavoc: :+1

app/assets/javascripts/client/views/directives/inventories/manage_participants.html
```diff
@@ -0,0 +1,14 @@
+<div class="row manage manage-participants plus-button" ng-click="showAddParticipants()">
```
MakotoTheKnight: Odd that this file is showing up now in this PR when it was last added into the repository...

spec/javascripts/controllers/inventories/AddInventoryUsersCtrlSpec.js
```diff
@@ -0,0 +1,52 @@
+(function() {
+  'use strict';
+  describe('Controller: AddInventoryUsersCtrl', function() {
```
MakotoTheKnight: Style:  You're describing a controller in your test, so there's no need to suffix it with `Ctrl` here.

app/assets/javascripts/client/inventories/AddInventoryUsersCtrl.js
```diff
@@ -0,0 +1,20 @@
...
+  AddInventoryUsersCtrl.$inject = ['$scope', 'InventoryInvitable', 'InventoryParticipant'];
+  function AddInventoryUsersCtrl($scope, InventoryInvitable, InventoryParticipant) {
+    var vm = this;
+    vm.reloadInvitables = function() {
+      vm.invitables = InventoryInvitable.list({ inventory_id: $scope.inventoryId });
+    }
+    vm.reloadInvitables();
+
+    vm.addUser = function(user) {
+      InventoryParticipant.create({ inventory_id: $scope.inventoryId }, { user_id: user.id }).$promise.then(function() {
```
MakotoTheKnight: I remember us discussing this the other day - it was either a matter of us passing this in as an attribute value or pulling it off of `$stateParams`.

At the time I said it was essentially the same, but when I read the code, I believe I prefer the `$stateParams` path better.

Simply put, we both agree that the only way this will be invoked is if it falls on a path parameter that contains the inventory ID.  Since that information is part of the route, we should be leveraging it.

It makes the code simpler in that:

 - We no longer need to include `$scope`
 - We use a more conventional and standard approach to pulling the path parameter

It makes the tests simpler in that we can supply our own take `$stateParams` without the need to run a digest cycle.

app/assets/javascripts/client/inventories/AddInventoryUsersCtrl.js
```diff
@@ -0,0 +1,20 @@
...
+  'use strict';
+
+  angular.module('PDRClient').controller('AddInventoryUsersCtrl', AddInventoryUsersCtrl);
+
+  AddInventoryUsersCtrl.$inject = ['$scope', 'InventoryInvitable', 'InventoryParticipant'];
+  function AddInventoryUsersCtrl($scope, InventoryInvitable, InventoryParticipant) {
+    var vm = this;
+    vm.reloadInvitables = function() {
+      vm.invitables = InventoryInvitable.list({ inventory_id: $scope.inventoryId });
+    }
```
MakotoTheKnight: Semicolon is missing here.

app/assets/javascripts/client/inventories/AddInventoryUsersCtrl.js
```diff
@@ -0,0 +1,20 @@
+(function() {
+  'use strict';
+
+  angular.module('PDRClient').controller('AddInventoryUsersCtrl', AddInventoryUsersCtrl);
+
+  AddInventoryUsersCtrl.$inject = ['$scope', 'InventoryInvitable', 'InventoryParticipant'];
+  function AddInventoryUsersCtrl($scope, InventoryInvitable, InventoryParticipant) {
+    var vm = this;
+    vm.reloadInvitables = function() {
```
MakotoTheKnight: `loadInvitables` may be a better name for this since we're still loading the invitables from the system, even after we hit the page for the first time.

app/assets/javascripts/client/inventories/addInventoryUsersLink.js
```diff
@@ -0,0 +1,18 @@
...
+
+  function addInventoryUsersLink() {
+    return {
+      restrict: 'E',
+      scope: {
+        inventoryId: '=inventoryId'
+      },
+      templateUrl: 'client/inventories/add_inventory_users_link.html',
+      controller: 'AddInventoryUsersLinkCtrl',
+      controllerAs: 'link',
```
MakotoTheKnight: I'd prefer a slightly longer `controllerAs` name to this as opposed to "link" which is open to a lot of interpretation.  Something like `inventoryUsersLink`.

app/controllers/v1/inventory_invitables_controller.rb
```diff
@@ -0,0 +1,21 @@
+class V1::InventoryInvitablesController < ApplicationController
+  before_action :authenticate_user!
+  authorize_actions_for :inventory
+
+  authority_actions index: 'read'
+  def index
+    member_users = inventory.members.select(:user_id).map(&:user_id)
+    @users = User.includes(:districts)
```
MakotoTheKnight: Is there a reason we can't use `joins`?  Unless there's some business logic constraint I believe we should, considering the query efficiency.

I ran this on staging with some sample data and saw these numbers.  The first query plan shows the `:includes`, and the second one is with `:joins`.

```
                                      QUERY PLAN
----------------------------------------------------------------------------------------
 Nested Loop  (cost=0.11..12.90 rows=1 width=450)
   ->  Nested Loop  (cost=0.06..8.84 rows=1 width=252)
         ->  Seq Scan on districts_users  (cost=0.00..4.77 rows=1 width=8)
               Filter: (district_id = 12)
         ->  Index Scan using users_pkey on users  (cost=0.06..4.06 rows=1 width=248)
               Index Cond: (id = districts_users.user_id)
               Filter: ((id <> 61) AND ((role)::text <> 'network_partner'::text))
   ->  Index Scan using districts_pkey on districts  (cost=0.06..4.06 rows=1 width=202)
         Index Cond: (id = 12)
(9 rows)

                                        QUERY PLAN
-------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.11..10.90 rows=1 width=248)
   ->  Nested Loop  (cost=0.06..8.84 rows=1 width=252)
         ->  Seq Scan on districts_users  (cost=0.00..4.77 rows=1 width=8)
               Filter: (district_id = 12)
         ->  Index Scan using users_pkey on users  (cost=0.06..4.06 rows=1 width=248)
               Index Cond: (id = districts_users.user_id)
               Filter: ((id <> 61) AND ((role)::text <> 'network_partner'::text))
   ->  Index Only Scan using districts_pkey on districts  (cost=0.06..2.06 rows=1 width=4)
         Index Cond: (id = 12)
(9 rows)
```

spec/javascripts/controllers/inventories/AddInventoryUsersCtrlSpec.js
```diff
@@ -0,0 +1,52 @@
...
+
+    it('retrieve initial invitables', function() {
+      $httpBackend.expectGET('/v1/inventories/4/invitables').respond([]);
+      $httpBackend.flush();
+    });
+
+    describe('#addUser', function() {
+      beforeEach(function() {
+        spyOn(subject, 'reloadInvitables').and.callThrough();
+        spyOn(InventoryParticipant, 'create').and.callFake(function(query, params) {
```
MakotoTheKnight: In this scenario, if you only want to test that the methods are invoked, you could still use `$httpBackend` to accomplish the request going through.

It'd be something similar to...

```$httpBackend.when('GET', '/v1/inventories/4/invitables').respond(200);```

...since you don't need the return value.

File: spec/javascripts/resources/InventoryInvitableSpec.js [hunk 14]
```diff
@@ -0,0 +1,20 @@
+describe('Resource: InventoryInvitable', function() {
```
MakotoTheKnight: This test isn't using the IIFE style.

spec/javascripts/resources/InventoryInvitableSpec.js
```diff
@@ -0,0 +1,20 @@
...
+  var subject, scope, $httpBackend;
+
+  beforeEach(module('PDRClient'));
```
MakotoTheKnight: You should look to move this inside of the same `beforeEach` function.

spec/javascripts/resources/InventoryInvitableSpec.js
```diff
@@ -0,0 +1,20 @@
...
+    $httpBackend = $injector.get('$httpBackend');
+  }));
+
+  it("returns all invitables users for an inventory", function(){
+    $httpBackend.expectGET('/v1/inventories/10/invitables').respond([1, 2]);
+    var users = subject.list({
+      inventory_id: 10
+    });
+    $httpBackend.flush();
+    expect(users.length).toEqual(2);
```
MakotoTheKnight: If I change what I get back from that `$httpBackend` call, I'll get back a different result, and this expectation will fail.  Essentially, `$httpBackend` is mocking out the response, so it's not indicative at all of what we're going to get back.

What would be a better test would be to ensure that it hits the right route.  I believe that the `expectGET` is sufficient to cover that case.

You would also want to be sure there are no outstanding expectations or outstanding requests after the test has run (through $httpBackend.verifyNoOutsandingExpectation()` and `$httpBackend.verifyNoOutsandingRequest()`).

spec/javascripts/resources/InventoryParticipantSpec.js
```diff
@@ -0,0 +1,21 @@
+describe('Resource: InventoryParticipant', function() {
```
MakotoTheKnight: This test isn't using the IIFE style.

spec/javascripts/resources/InventoryParticipantSpec.js
```diff
@@ -0,0 +1,21 @@
...
+
+  beforeEach(module('PDRClient'));
+
+  beforeEach(inject(function($resource, $rootScope, InventoryParticipant, $injector) {
+    scope = $rootScope.$new();
+    subject = InventoryParticipant;
+    $httpBackend = $injector.get('$httpBackend');
+  }));
+
+  it("creates participant", function(){
```
MakotoTheKnight: Similar to the previous resource test, this seems to be the right approach, but what you want to do is rename this expectation.  We would expect that the create method is called with the parameters we expect it to.

bithavoc:
@MakotoTheKnight added behavior expectations to issue #83

MakotoTheKnight:
Based on the expectations and what I had observed, I'm believing that this PR is good to go.  We're going to want to add information in #144 that we need to wire this functionality into that page.

:shipit:

PR #166, opened 2016-03-30
Closes #132:  Separate participant management

### LINKED ISSUE #132: Add additional buttons for adding participants and separate modals

MakotoTheKnight:

 - This affects the assessment creation and dashboard pages.
 - This adds in the new button style that was requested by UX.  Be sure that we haven't stomped out any other designs anywhere else (notably with the plus button).
 - The behavior of the modals should be consistent with whatever is currently in staging.

---

## A bit on the modals architecture

The modals themselves are broken out in this manner.

 - The modals are driven by their own controller, which includes facilities to submit data and update data as necessary.
 - That controller is attached to a directive (ending in `Modal`) which encapsulates the entirety of the modal itself.
 - That is then leveraged in another controller which drives the actual creation of the modal through the `$modal` service.
 - That controller is attached to a directive which contains the button to activate the modal.

The hierarchy then becomes:

    buttonDirective -> ButtonController
                                   |
                                   |
                                   v
                       buttonModalDirective -> ButtonModalController

It is the case that the `ButtonController` is responsible for closing the `buttonModalDirective`, and this is accomplished with `$scope.$emit` and unique events across all three events.  This approach also decouples the creation of the modal from the modal itself, should we wish to ever move them to a view that isn't a modal.


juliosantos:
This may be outside of the scope of this PR, but it bothers me slightly that loading some content happens after the modals are open. For example, clicking "Add New Participants" invokes a modal which is empty when it loads, and populates only a few seconds later.
Other than that, LGTM.

MakotoTheKnight:
@juliosantos:  I think the likely reason for that is due to the fact that data is loading in that modal.  An enhancement could be to add a loading throbber so that it's more apparent as to what is going on.

juliosantos:
@MakotoTheKnight: Another alternative is to have the routes `require` the data that will likely be necessary for its view.

PR #174, opened 2016-04-01
Closes #163: Migrate over all inventory string fields to text

### LINKED ISSUE #163: Database: migrate all instances of string to text
A `string` in PG is limited to 255 characters. Space being cheap and this not being presumably a project requirement, we should migrate all `string` into `text` to avoid PG errors or the need for backend (and maybe frontend) validations to guard against that.

MakotoTheKnight:

 - This only changes the inventory questionnaire-related models' columns over from string to text.
 - No data should be destroyed with this migration, but that should be double checked.

### DIFF
--- a/app/models/data_access_question.rb
+++ b/app/models/data_access_question.rb
@@ -3,11 +3,11 @@
 # Table name: data_access_questions
 #
 #  id                   :integer          not null, primary key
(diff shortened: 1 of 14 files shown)


bithavoc:
LGTM

PR #189, opened 2016-04-06
Hide inventory nav to allow deployment to production

### LINKED ISSUE #157: Feature Toggle support for deploys
We will require a feature flag to deploy what we've got on staging to production.  Ideally, this should take the form of an environment flag which guards parts of our UI from a code path we are not ready to expose.

- [ ] Create table to capture feature name and if it's enabled, indexed on name (no ID)
- [ ] Create endpoint to leverage table
- [ ] Create V1 API controller for feature toggles
- [ ] Create seed data for staging and production (seed desired feature ON in staging, OFF in production)
- [ ] Create Angular resource to determine if feature is eligible to be seen, with caching
- [ ] Hide feature using `ng-if`

seanperkins:
Closes #157 

Hiding inventory navigation without disabling inventory for now.


bithavoc:
Looks like this cover most of the cases to access inventory in the web app

Issue #183, opened 2016-04-04
Create /inventories/:id/products resource


MakotoTheKnight:
For tasks #179 and #165, we require an endpoint to pull back a listing of product entries.

Data returned should be:

 - id
 - title
 - vendor
 - purpose (correlates to "Used for")
 - hosting (correlates to "Where is the data hosted?")
 - attached_to_other_products (boolean field)
 - single_sign_on
 - usage (represents staff usage)

bithavoc:
I'll be researching what angular datatable component(likely to be http://l-lin.github.io/angular-datatables) we need and how the API would fit the needs of the data table backend.

bithavoc:
@MakotoTheKnight 
in **Data returned should be**, is `title` supposed to be `product_name`?

MakotoTheKnight:
We can likely call the resource `product_name`.  This was meant to correspond to the original version of these mockups found at https://drive.google.com/a/mobility-labs.com/file/d/0B-nCEm5iPUjiOG9GaGZQSWs5QkE/view.

Issue #179, opened 2016-04-04
Edit Inventory Page: Product Listing


MakotoTheKnight:
Wire in the product listing.

 - Add filters and search bar
 - Add product listing (including information for product name, vendor, and "used for")

bithavoc:
I'm noticing in the filters there is an option to filter by `Vendor`, it looks like a checkbox:

![](https://cloud.githubusercontent.com/assets/219289/14255058/07982b64-fa58-11e5-8cd6-9b5d5532218f.png)

However, in https://github.com/MobilityLabs/pdredesign-server/issues/104 `Vendor` is a text input:

![](https://cloud.githubusercontent.com/assets/219289/14255096/2e4bf1aa-fa58-11e5-86c2-ff9a88a6bd17.png)

Does this mean the filter/option to be layout out in https://github.com/MobilityLabs/pdredesign-server/issues/178 should be a textbox? If this is true, the [design attached to google drive in #177](https://github.com/MobilityLabs/pdredesign-server/issues/177#issuecomment-205347903) should be updated to show how a filter with text input would look like before we start working in #178

bithavoc:
@seanperkins just clarified me that these are not filtering options, they are instead options to modify which columns are being presented in the data table.

MakotoTheKnight:
@bithavoc:  Okay.  That makes defining what needs to be brought back with #183 a bit clearer.

bithavoc:
@MakotoTheKnight do you think we need server side support to prevent all the remaining columns not being presented in the list from being fetched as part of the results or is this something we can address when it becomes a real perfomance issue?

MakotoTheKnight:
@bithavoc My thought is that we can bring back all of these rows and filter them based on what we select in the view.

seanperkins:
I would like to see what we are looking at performance wise if we just default to all fields in the endpoint initially. If we see that it becomes an issue with 30-50 inventory and data items, add additional filtering, otherwise, roll with what we have.

seanperkins:
Looks like we are all in agreement to roll with all fields initially

MakotoTheKnight:
After discussion, the product listing will not include the field for "Usage Data from Vendor".

HeyCat:
![inventory_no_entries](https://cloud.githubusercontent.com/assets/5464910/14284550/049f6e72-fb16-11e5-99be-d8419fd39daa.png)
[...]

PR #191, opened 2016-04-07
#178 inventory edit base layout

### LINKED ISSUE #178: Edit Inventory Page:  Base Layout
Based on #177, we are going to be creating the base structure of this page.

There should be:

 - Non-functional buttons (Mark Inventory Complete, Save and Exit Inventory, Add product entry, Add data entry)
 - Correct copy for page
 - Learning Questions will list in its modal as opposed to the sidebar

bithavoc:
Product and Data entry lists is being intentionally left out of this PR because the implementation leaks to [products list server-side](https://github.com/MobilityLabs/pdredesign-server/issues/183) and [products list UI](https://github.com/MobilityLabs/pdredesign-server/issues/179)

Closes #178


app/controllers/v1/inventories_controller.rb
```diff
@@ -1,12 +1,20 @@
...
-
   before_action :authenticate_user!
 
   def index
     @inventories = Inventory.joins('LEFT OUTER JOIN inventory_members ON inventory_members.inventory_id = inventories.id')
                        .where('inventories.owner_id = ? OR inventory_members.user_id = ?', current_user.id, current_user.id)
   end
 
+  def show
+    @inventory = Inventory.where(id: params[:id]).first
```
MakotoTheKnight: We might collide here...I've got some work that does something very similar, although I may wind up accepting this since it behaves better when nothing comes back.

One note to make:  Is there a reason we would prefer `Inventory.where(...)` as opposed to `Inventory.find(...)`?

MakotoTheKnight:
LGTM :shipit:

PR #201, opened 2016-04-11
#106 user can fill data entry

### LINKED ISSUE #106: A User Can Fill Out a Data Entry

juliosantos:
closes #106


app/assets/javascripts/client/inventories/InventoryDataEntryCtrl.js
```diff
@@ -0,0 +1,26 @@
...
+    '$scope',
+    '$modal'
+  ];
+
+  function InventoryDataEntryCtrl($scope, $modal) {
+    var vm = this;
+
+    vm.showInventoryDataEntryModal = function() {
+      vm.modalInstance = $modal.open({
+        template: '<inventory-data-entry-modal inventory="inventory" resource="resource"></inventory-data-entry-modal>',
```
MakotoTheKnight: Where is it pulling the inventory and resource from?  I'm under the impression that those are strings, but I may be wrong on that.

MakotoTheKnight:
Two things (neither are critical):

 - There are a lot of annotation-driven changes which add a limit of 255 in the spec.  It's likely not important since it pulls from local settings, but we're going to want to address that at some point in the future.
 - We're not doing anything right now with the `data_storage` column because it relates to products, right?  We should capture a task to finish that piece off.

app/assets/javascripts/client/inventories/InventoryDataEntryCtrl.js
juliosantos: It's pulling it from the parent scope (`inventoryDataEntry`'s: `<inventory-data-entry inventory="inventory" resource="resource"></inventory-data-entry>`). This is the same method I used for product entries, which in turn was adapted from your at https://github.com/MobilityLabs/pdredesign-server/blob/5ae5c17ed45ef0ee97ade3eed2b600186f4e3eeb/app/assets/javascripts/client/permissions/AddNewParticipantsCtrl.js.

juliosantos:
> We should capture a task to finish that piece off.
Indeed, https://waffle.io/MobilityLabs/pdredesign-server/cards/570f5fe473fcb1950097547f

MakotoTheKnight:
I believe we're good with this PR.  Go ahead and rebase it against staging once more and merge it in.  :shipit:

bithavoc:
LGTM, go go go!

bithavoc:
Once this branch is rebased, it will also unblock https://github.com/MobilityLabs/pdredesign-server/pull/206

Issue #200, opened 2016-04-11
User cannot log into account after being invited to an inventory


MakotoTheKnight:
Related to #144, if a facilitator attempts to invite a user to an inventory via email, and the user attempts to create an account or sign in to the account, they are rejected due to the account already having been created.

The flow for this should be similar to how invites for assessments work; allow the invited user to change details of their account when they attempt to log in with it.

Issue #193, opened 2016-04-08
Add support for messages to Inventory model


MakotoTheKnight:
This came up during #144:  We need to support messages in the same manner that Assessments do.

 - [x] Make the existing `messages` table polymorphic and migrate existing messages to the new polymorphic scheme (all messages in the system will be by default 'Assessment' type)
 - [x] Create the inventory_helper for the default inventory message, analogous to assessment_helper.rb
 - [x] Incorporate the message entry into the JSON payload for inventories
 - [x] Add new copy to site
 - [x] Ensure invite emails can be sent from the InventoriesController

MakotoTheKnight:
@candyj0311: This requires copy.  The existing copy for an assessment reads as thus:

 > The goal of this assessment is to discuss professional development and determine areas of strength or limitations within your district. To participate along with your colleagues, you must first register with the PD Readiness Assessment Tool. Once registered, you’ll be able to answer the questions and have on-going access to the assessment discussions.

MakotoTheKnight:
#144 depends on this component for it to be considered complete. We presently do not have a way to send out inventory emails and determine completion without it.

PR #202, opened 2016-04-11
Closes #144: Create Inventory page

### LINKED ISSUE #144: Product Inventory:  Create Page

MakotoTheKnight:
This PR is ready.  Notes:

 - Issues with the invite email, which includes sending the email and its associated copy are captured in #193.
 - A user invited to the inventory through this process cannot finish the registration process - this is captured in #200.

Regression points:

 - User management (invites, adding participants, and managing permissions) across both assessments create page and dashboard
 - Learning questions on dashboard and create page for assessments


bithavoc:
Functionality-wise this PR it seems to do the trick, except for one thing: Inviting a new participant doesn't automatically refresh the list of participants, I had to do F5-like refresh to see the new participant in the list.

app/assets/javascripts/client/common_create/CreateService.js
```diff
@@ -6,16 +6,35 @@
...
-    'Assessment'
+    '$stateParams',
+    'Assessment',
+    'Inventory',
+    'Participant',
+    'InventoryParticipant'
   ];
 
-  function CreateService($window, $location, Assessment) {
+  function CreateService($window, $location, $stateParams, Assessment, Inventory, Participant, InventoryParticipant) {
```
bithavoc: Is this a service to create tools? how about naming it `ToolCreateService`?

app/assets/javascripts/client/common_create/ToolDetailsCtrl.js
```diff
@@ -37,7 +37,11 @@
 
     vm.save = function(entity) {
       CreateService.loadDistrict(vm.district);
-      CreateService.saveAssessment(entity);
+      if(CreateService.context === 'inventory') {
```
bithavoc: I think this functionality is envying `CreateService`, if we pass the entity to `CreateService.save` then the service should be able to figure it out.

app/assets/javascripts/client/common_create/CreateService.js
MakotoTheKnight: I hadn't settled on the "tool" naming convention until a fair bit after this service had been created.  I'm on board for renaming it, but it is something that we could tackle in subsequent cleanup work?

app/assets/javascripts/client/common_create/ToolDetailsCtrl.js
MakotoTheKnight: This is probably a good change to make.  It also jives with pushing the creation and API logic out of the controllers and into services.

config/routes.rb
```diff
@@ -78,13 +78,20 @@
...
       resource :permissions, controller: 'inventory_permissions', only: [:show, :update]
-      resources :participants, controller: 'inventory_participants', only: [:create, :destroy]
+      resources :participants, controller: 'inventory_participants', only: [:create, :destroy, :index]
+        get 'participants/all', to: 'inventory_participants#all'
       resources :invitables, controller: 'inventory_invitables', only: [:index]
       resources :product_entries, only: [:index, :show, :create, :update]
+
+      resources :learning_questions, only: [:index, :create, :update, :destroy] do
+        collection do
+          get 'exists', to: 'learning_questions#exists'
```
bithavoc: This endpoint may need some clarification on why doesn't `#show` satisfy our needs, also since it's part of the collection `exists?` doesn't really answer any specific question.

bithavoc:
LGTM

config/routes.rb
MakotoTheKnight: It's a smaller payload - all I want to verify is if the entity exists before I do anything else with it.  It's used for determining if we wish to open the modal for a user to encourage them to create learning questions.  It can be found here:  https://github.com/MobilityLabs/pdredesign-server/blob/staging/app/assets/javascripts/client/controllers/responses/ResponseCtrl.js

I don't use the `show` since I don't want to bring back more data than I care about on the request.  All I need is the 200 and if I can save the bandwidth, I'd like to.

MakotoTheKnight:
I noticed that this doesn't index the new polymorphic columns correctly, so I'll make that adjustment alongside the review findings.

MakotoTheKnight:
@bithavoc:  I've made the tweaks to this PR.  If you've got no further issues I can merge it into staging tomorrow.

bithavoc:
LGTM

PR #204, opened 2016-04-11
#179 #183 Inventory products list

### LINKED ISSUE #179: Edit Inventory Page: Product Listing

### LINKED ISSUE #183: Create /inventories/:id/products resource

bithavoc:
Closes #179
Closes #183

### DIFF
--- a/app/assets/javascripts/client/common_list/entry_view.html
+++ b/app/assets/javascripts/client/common_list/entry_view.html
@@ -58,11 +58,6 @@
         <span class="stat">{{entity.participant_count}}</span>
         <span class="category">Completed</span>
         <span class="stat">{{entity.completed_responses}}</span>
-        <!-- <add-inventory-users-link inventory-id="entity.id"></add-inventory-users-link> -->
-        <!-- <manage-inventory-permissions-link inventory-id="entity.id"></manage-inventory-permissions-link> -->
-        <!-- <invite-inventory-user-link inventory-id="entity.id"></invite-inventory-user-link> -->
-        <product-entry ng-if="isAssessment === 'false'" inventory="entity"></product-entry>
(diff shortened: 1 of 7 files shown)


juliosantos:
This PR is included in https://github.com/MobilityLabs/pdredesign-server/pull/206, right? My comments there should apply here too.

MakotoTheKnight:
If the screen's just the right size, we could run into this bleeding over the edge:

image

I'd recommend that we leverage some of Bootstrap's `table` classes.

bithavoc:
@MakotoTheKnight added the responsive style to the table, it worked.

as for the sidebar buttons, we can solve this by incrementing the size of the sidebar in `app/views/pdr_client/layouts/application.html.erb`:

from `col-md-9` to `col-md-8`
and `col-md-3` to `col-md-4`.



thoughts?

bithavoc:
@MakotoTheKnight @juliosantos please take another look at this PR

MakotoTheKnight:
We're still missing an empty state here:

image

I would imagine that we'd want to prevent the user from hiding everything except the column we're searching on in this view.  @HeyCat:  thoughts on that?

MakotoTheKnight:
Alright, this looks good to me.

I've played around a bit with the search and it seems that we can search on any column, which may cause some ambiguity for users searching for "Alpha Product" and having a vendor of their product actually being called "Al".

That, I'm not so bothered with.  We can roll out with this functionality.

:shipit:

PR #207, opened 2016-04-12
188 inventory list should not contain duplicates

juliosantos:
closes #188

PR #208, opened 2016-04-12
#207 188 inventory list should not contain duplicates

### LINKED ISSUE #188: An inventory with multiple members should not appear multiple times in the list of inventories

juliosantos:
Closes #188

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -81,6 +81,7 @@ class User < ActiveRecord::Base
 
   def inventories
     Inventory.
+        distinct.
         joins('LEFT OUTER JOIN inventory_members ON inventory_members.inventory_id = inventories.id').
         where('inventory_members.user_id = ?', self.id)
   end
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -168,4 +168,23 @@
       end
     end
   end
+
+  describe '#inventories' do
+    let(:user) { FactoryGirl.create(:user, first_name: 'John', last_name: 'Doe') }
+    context 'with both a first and last name' do
```
MakotoTheKnight: The context here is wrong, but the test here is correct.

Main things to note:  we could probably move the creation of those entities into two separate contexts; one test covering when the user is the owner, and one when the user isn't.  But this satisfies the requirements.

This specific spec be addressed when we go back to spruce up the code base.

app/models/user.rb
```diff
@@ -81,6 +81,7 @@ def ensure_district(district:)
 
   def inventories
     Inventory.
+        distinct.
```
MakotoTheKnight: About what I had in mind. :+1:

MakotoTheKnight:
Nothing here is blocking this particular piece.  :shipit:

PR #212, opened 2016-04-13
165 product entries connectivity

### LINKED ISSUE #165: Product Entries: Connectivity Section
This was moved out of #104 for simplicity, and also because the search is probably generifiable.

juliosantos:
closes #165 
I chose not to use the data tables for the connectivity section, because:
* it’s a list, not a table
* the only common functionality would be search

Instead, I created a `<searchable-list></searchable-list>` directive with transclusion which wraps around whatever list you like, and creates a search interface for it.

I'm not 100% convinced that the `district_product_entries` action belongs in `InventoriesController`, or that the `for_district` scope really sits well inside the `ProductEntry` model, so please do feel free to suggest a better place for those.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160404201732) do
+ActiveRecord::Schema.define(version: 20160412111252) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -19,10 +19,10 @@ ActiveRecord::Schema.define(version: 20160404201732) do
   create_table "access_requests", force: :cascade do |t|
     t.integer  "assessment_id"
     t.integer  "user_id"
-    t.string   "roles",         default: [], array: true
+    t.string   "roles",                     default: [], array: true
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.string   "token"
+    t.string   "token",         limit: 255
   end
 
   create_table "answers", force: :cascade do |t|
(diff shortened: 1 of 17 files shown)


db/schema.rb
```diff
@@ -11,18 +11,18 @@
...
 
   create_table "access_requests", force: :cascade do |t|
     t.integer  "assessment_id"
     t.integer  "user_id"
-    t.string   "roles",         default: [], array: true
+    t.string   "roles",                     default: [], array: true
     t.datetime "created_at"
     t.datetime "updated_at"
-    t.string   "token"
+    t.string   "token",         limit: 255
```
MakotoTheKnight: Be sure that this limit is coming from a migration that's already defined.  It's touching a lot of tables that aren't related to this work specifically.

MakotoTheKnight:
A couple of notes:

- Nothing in the connectivity column [appears to] tie[s] those IDs back to a `ProductEntry` that exists.  If we do run into the scenario in which an entry is deleted (soft or otherwise), we don't want to maintain that association.
 - The data being brought back is more than I thought we'd need on this endpoint, but for now I'll live with it.  We want to bring back only the stuff we care about in this scenario; I was envisioning just an ID and the name of the product, but what we have now is fine.

db/schema.rb
juliosantos: Could be an issue with environment differences. Strings are `limit: 255` by default in postgresql, so I'm not sure it's a problem.

juliosantos:
> Nothing in the connectivity column [appears to] tie[s] those IDs back to a ProductEntry that exists. If we do run into the scenario in which an entry is deleted (soft or otherwise), we don't want to maintain that association.

I didn't think this was an issue because the direction was to «Rework the technical_questions table to accept an `integer[]` instead of `text`». There's two ways I see to go about this:

1. Using a join table.
2. Using:
  * a `before_save` callback on `TechnicalQuestion` to clean the `connectivity` array out of any nonexistent `ProductEntry` `id`s
  * an `after_destroy` callback on `ProductEntry` which takes the destroyed `id` and wipes it out of all `connectivity` arrays

1 is more Rails-y, and 2 is more brittle. But 2 might be quicker to implement. @MakotoTheKnight: thoughts?

MakotoTheKnight:
We'll need to decide what behavior we want when we "delete" an inventory item; would we be removing it from the connectivity table as well, or would we keep it there but hide it in the listing?  It's probably not something we should tackle in this task.

All in all, this looks good to me.  Rebase against staging and merge it in. :shipit:

PR #216, opened 2016-04-13
Closes #193: Add messages to inventories

### LINKED ISSUE #193: Add support for messages to Inventory model

MakotoTheKnight:

 - This builds on top of work done for #144 in that we complete the flow for inviting inventory participants.
 - Persistence should still behave as it did before.
 - Copy will likely need to be inspected to ensure that the emails and the inventory copy page match.  There shouldn't be any significant difference between the two.
 - This will need to be merged _after_ #202 is merged in, although these two can be reviewed at the same time.


app/assets/javascripts/client/common_create/CreateService.js
```diff
@@ -6,16 +6,35 @@
...
+    this.extractCurrentDistrict = function(user, entity) {
+      var result = user.districts[0];
+      for (var i = 0; i < user.districts.length; i++) {
+        if (entity.district_id === user.districts[i].id) {
+          result = user.districts[i];
+          break;
+        }
+      }
+      return result;
+    };
```
juliosantos: Underscore can abstract this complexity away for us: `return _.findWhere(user.districts, {id: entity.district_id}) || user.districts[0]`

app/assets/javascripts/client/common_create/CreateService.js
```diff
@@ -80,5 +99,103 @@
...
+
+    this.save = function(entity) {
+      if(service.context === 'assessment') {
+        this.saveAssessment(entity);
+      } else if(service.context === 'inventory') {
+        this.saveInventory(entity);
+      }
+    };
+
+    this.loadParticipants = function() {
```
juliosantos: This method and the ones below might be clearer in terms of readibility if broken down into components, much like `save` does it.
```javascript
this.loadParticipants = function () {
  if(this.isAssessmentContext()) {
    return loadAssessmentParticipants();
  } else if(this.isInventoryContext()) {
    return loadInventoryParticipants();
  }
}

this.loadAssessmentParticipants = function () {
  return Participant.query({assessment_id: $stateParams.id});
}

this.loadInventoryParticipants = function () {
  return InventoryParticipant.query({inventory_id: $stateParams.id});
}
```

Using these utilities might also be beneficial:
```javascript
this.isAssessmentContext = function () { return service.context === 'assessment'; };
this.isInventoryContext = function () { return service.context === 'inventory'; };
```

app/models/message.rb
```diff
@@ -6,20 +6,21 @@
...
 class Message < ActiveRecord::Base
-	belongs_to :assessment
+	belongs_to :tool, polymorphic: true
 	MAX_LENGTH_TEASER = 220
 
 	def teaser
 	  return content unless content
     return content if content.length <= MAX_LENGTH_TEASER
-    return "#{content[0..MAX_LENGTH_TEASER-1]}..."
+    "#{content[0..MAX_LENGTH_TEASER-1]}..."
```
juliosantos: We could use `truncate` here:
```ruby
content.truncate(MAX_LENGTH_TEASER, omission: "...")
```
This way we don't need both lines 23 and 24.

Also: `...` is 3 characters, so the end result will be longer than 220.

spec/models/message_spec.rb
```diff
@@ -16,11 +16,45 @@
...
+    context 'when printing a string larger than 220 characters' do
+      before(:each) do
+        message.content = Faker::Lorem.characters(221)
+      end
+
+      it 'limits the string to 223 characters' do
+        expect(message.teaser.size).to eq 223
+      end
+
+      it 'adds an ellipsis at the end' do
```
juliosantos: Might be nitpicking, but an ellipsis is `…`. 
`...` is three dots.

spec/models/message_spec.rb
```diff
@@ -6,21 +6,56 @@
...
```
juliosantos: Might be nitpicking, but an ellipsis is `…`.
`...` is just three dots.

app/assets/javascripts/client/common_create/CreateService.js
juliosantos: Also: do we want to return `.$promise` here, like everywhere else?

app/models/message.rb
MakotoTheKnight: This is the way the original code behaved.  It would truncate what existed in the database down to 220 characters, then tack on an additional 3.

app/controllers/v1/inventory_participants_controller.rb
```diff
@@ -18,8 +23,15 @@ def destroy
...
   end
 
-  protected
+  def all
+    district = inventory.district
+    participant_ids = inventory.members.pluck(:user_id)
+    @users = User.includes(:districts).where(districts: {id: district.id}).where.not(id: participant_ids)
+                 .reject { |u| u.role.to_sym == :network_partner }
+  end
+  authority_actions all: 'update'
```
juliosantos: Are we sure about this? Even if the permissions for getting this list and updating a record are the same, I find this line confusing.

MakotoTheKnight: This is analogous to how `ParticipantsController` presently behaves at https://github.com/MobilityLabs/pdredesign-server/blob/master/app/controllers/v1/participants_controller.rb#L29.

spec/javascripts/inventories/AddInventoryUsersCtrlSpec.js
```diff
@@ -1,47 +1,54 @@
...
         $httpBackend.when('GET', '/v1/inventories/4/invitables').respond([]);
-      })
+      });
     });
 
-    it('retrieve initial invitables', function() {
+    it('retrieves initial invitables', function() {
       $httpBackend.expectGET('/v1/inventories/4/invitables').respond([]);
-      $httpBackend.flush();
+      expect($httpBackend.flush).not.toThrow();
```
juliosantos: What exactly is this doing?

app/controllers/v1/inventory_participants_controller.rb
bithavoc: are these the users we can invite from the district?

bithavoc: I'm afraid we already have [inventory_invitables_controller](https://github.com/MobilityLabs/pdredesign-server/blob/staging/app/controllers/v1/inventory_invitables_controller.rb#L7) for this purpose

spec/javascripts/learning_questions/LearningQuestionFormCtrlSpec.js
```diff
@@ -0,0 +1,150 @@
...
+
+    describe('#validate', function() {
+      var data = 'This is some data';
+
+      beforeEach(function() {
+        spyOn(LearningQuestionService, 'validate');
+        controller.validate(data);
+      });
+
+      it('delegates to LearningQuestionService', function() {
```
juliosantos: I've seen this pattern elsewhere, but it seems to me that all this is testing is code flow. If we change the implementation but achieve the same results, this test would fail. Would we want it to?

MakotoTheKnight: From the view of this controller, all it's doing is delegating to the service which does the heavy lifting.  If we change what it's delegating to, that is a valid test failure.

If the implementation changes, I would expect the `LearningQuestionService`'s tests to catch any breakage, since that service is doing the main body of work.

app/assets/javascripts/client/common_create/CreateService.js
```diff
@@ -80,5 +85,45 @@
...
+    };
+
+    this.saveInventory = function(inventory, assign) {
+      if (inventory.name === '') {
+        this.emitError('Inventory needs a name!');
+        return;
+      }
+
+      inventory.district_id = this.district.id;
+      service.toggleSavingState();
```
juliosantos: Would it be more consistent to either use `this` or, preferably, `service` in all instances?

MakotoTheKnight: I wouldn't be opposed to it.  We can address this in a style cleanup.

app/mailers/inventories_mailer.rb
```diff
@@ -0,0 +1,29 @@
...
+  def reminder(inventory, message, participant)
+    @participant_name = participant.user.first_name
+    @facilitator_name = inventory.user.name
+    @inventory_name = inventory.name
+    @inventory_link = inventory_url(inventory.id)
+    @district_name = inventory.district.name
+    @deadline = inventory.due_date.strftime("%B %d, %Y")
+    @message = message.html_safe
+
+    mail(subject: 'Inventory Reminder', to: participant.user.email)
```
juliosantos: no need for html+text here?

app/models/inventory.rb
```diff
@@ -9,15 +9,24 @@
...
 
 class Inventory < ActiveRecord::Base
   include Authority::Abilities
+
+  default_scope { order(created_at: :desc) }
+
   has_many :product_entries
   has_many :data_entries
   has_many :access_requests, class_name: 'InventoryAccessRequest'
+  has_many :messages, foreign_key: :tool_id
```
juliosantos: The `Assessment` model appends `dependent: :destroy` to this. Are we sure we don't want it here too?

juliosantos:
If you don't select a value from the _Set Permissions Level_ dropdown, the new `InventoryMember`'s role is not set. This means they won't be emailed, since `inventory.participants` (in `AllInventoryParticipantsNotificationWorker#l6` will be an empty array because:
```ruby

```

MakotoTheKnight:
I think what we'll do is default that to 'participant', since:

 - It's the weakest level of permission
 - It can be changed at a later stage

MakotoTheKnight:
@juliosantos:  So I took a deeper dive at this, and figured out the specific reason why this behaves this way.  Long story short: this is also how assessments behave on the form; it does not select a default permission level.

While I was looking into this, this actually did expose a regression in which we passed `role` all the way through to this form, so I went ahead and restored that.

My gut instinct on this is to default to participant on this, but we are going to want to revisit what this means on a more global scale. To mind, I haven't observed anyone having trouble with the existing assessment flow.

MakotoTheKnight:
@juliosantos @bithavoc:  If either of you want to smoke test the flow and ensure that the emails are sent off, that'd be great.  Since we default to a value I'm fairly confident it works now.

app/assets/javascripts/client/permissions/InviteUserCtrl.js
```diff
@@ -13,6 +13,7 @@
     var vm = this;
 
     vm.showInviteUserModal = function() {
+      console.log($scope.role);
```
juliosantos: Whoops.

juliosantos:
I noticed that the users _added_ to an inventory would see its `message` in the email, but the _invited_ ones would not. Given that the email template included `@message`, I assumed it was an oversight on `InventoryInvitationMailer` and fixed it. Hope I'm not overreaching by committing to this PR.

MakotoTheKnight:
@bithavoc Mind taking a look?  I have a feeling this may tie into work you're doing next.

bithavoc:
Inviting a new participant doesn't refresh the list of participants in the assign page upon closing the **Add New Participant** modal.

bithavoc:
@MakotoTheKnight where can I find the original copy?

MakotoTheKnight:
@bithavoc: Would you give this one more look?

bithavoc:
LGTM :shipit:

PR #221, opened 2016-04-14
Closes #126 send invitations

### LINKED ISSUE #126: Resend Inventory invitations email
This would involve moving our logic from the `send_invite` flag in `invitations_controller` to a new controller specific for this purpose

bithavoc:


juliosantos:
I'm not sure how this PR jives with https://github.com/MobilityLabs/pdredesign-server/pull/216.
I suppose that `Inventory#send_invites` has its place, but invoking that at inventory could be a duplication of @MakotoTheKnight's work.

MakotoTheKnight:
After discussing it with the team, we're going to close this PR without merging it.  We believe there's overlap with existing work in #216.

PR #224, opened 2016-04-15
Closes #114 - export inventory data to CSV

### LINKED ISSUE #114: Users Can Export Inventory Data to CSV
Once an inventory is completed, facilitators and participants can download the inventory data to a .csv file. We already have this functionality for the Readiness Assessment.

juliosantos:
I'm not too happy with the code duplication, but I'm also not sure we should invest in building a higher-level abstraction around this.


MakotoTheKnight:
This looks good to me.  :shipit:

Issue #196, opened 2016-04-11
New Analysis Modal (No Inventory)


candyj0311:
If a user does not have a completed inventory, this modal should appear when they begin the process of creating an Analysis.

candyj0311:
![image](https://cloud.githubusercontent.com/assets/10750162/14427547/599bb1d6-ffc3-11e5-809e-4a63d5a838da.png)

candyj0311:
I'll update the lorem ipsum with real copy.

candyj0311:
Copy: The Data & Tech Analysis needs to be linked to a Data & Tech Inventory to be most effective. If you've already built an inventory, check to make sure you are a participant or facilitator of that inventory in order to create an analysis. You can also start a new inventory by clicking below.

juliosantos:
@candyj0311 I'll assume the "Save" button should have been dropped from the mockup, let me know if I'm wrong.

candyj0311:
@juliosantos: Yes, that Save button doesn't do anything. @heycatrobinson LMK if you want any changes to the modal beyond dropping the unnecessary Save. I can't think of another modal that behaves like this, so I don't think there's a standard to maintain around buttons to display.

HeyCat:
![new-analysis-no-inventory](https://cloud.githubusercontent.com/assets/5464910/14644914/ae126be2-0621-11e6-9e1a-d103d2124c6f.png)

PR #238, opened 2016-04-18
197 new analysis modal

### LINKED ISSUE #197: New Analysis Modal (User Has An Inventory)
If the user does have 1 or more inventories, this modal should appear:

image

Copy: Facilitate the Data & Tech Analysis process for your district by creating a new analysis and selecting an inventory to connect to it. Click here to view current analyses for your district or to participate in an existing analysis.

juliosantos:
Closes #197

### DIFF
--- a/app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
+++ b/app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
@@ -6,22 +6,35 @@
 
   AnalysisModalCtrl.$inject = [
     '$scope',
-    'Analysis'
+    'Analysis',
+    'Inventory'
   ];
 
-  function AnalysisModalCtrl($scope, Analysis) {
+  function AnalysisModalCtrl($scope, Analysis, Inventory) {
     var vm = this;
 
     vm.closeModal = function() {
       $scope.$emit('close-analysis-modal');
     };
 
+    vm.updateData = function () {
+      return vm.updateInventories();
+    };
+
+    vm.updateInventories = function () {
+      return Inventory.query().$promise.then(function (response) {
+        vm.inventories = response.inventories;
+
+        // ensure a district is always selected
+        vm.analysis.inventory_id = vm.inventories[0].id;
+      });
+    };
+
+
     vm.analysis = {};
 
     vm.save = function () {
-      Analysis.create({
-        inventory_id: vm.inventory.id
-      }, vm.analysis)
+      Analysis.create(null, vm.analysis)
           .$promise
           .then(function (productEntry){
             vm.closeModal();
--- a/app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
+++ b/app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
@@ -1,0 +1,32 @@
+(function() {
+  'use strict';
+
+  angular.module('PDRClient')
+      .controller('AnalysisModalCtrl', AnalysisModalCtrl);
+
+  AnalysisModalCtrl.$inject = [
+    '$scope',
+    'Analysis'
+  ];
+
+  function AnalysisModalCtrl($scope, Analysis) {
+    var vm = this;
+
+    vm.closeModal = function() {
+      $scope.$emit('close-analysis-modal');
+    };
+
+    vm.analysis = {};
+
+    vm.save = function () {
+      Analysis.create({
+        inventory_id: vm.inventory.id
+      }, vm.analysis)
+          .$promise
+          .then(function (productEntry){
+            vm.closeModal();
+          }, function (response) {
+          });
+    };
+  }
+})();
--- a/app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
+++ b/app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
@@ -12,6 +12,8 @@
 
   function AnalysisModalCtrl($scope, Analysis, Inventory) {
     var vm = this;
+    vm.analysis = {};
+    vm.alerts = [];
 
     vm.closeModal = function() {
       $scope.$emit('close-analysis-modal');
@@ -30,8 +32,14 @@
       });
     };
 
+    vm.addAlert = function(message) {
+      vm.alerts.push({type: 'danger', msg: message});
+    };
 
-    vm.analysis = {};
+    vm.closeAlert = function(index) {
+      vm.alerts.splice(index, 1);
+    };
+
 
     vm.save = function () {
       Analysis.create(null, vm.analysis)
@@ -39,6 +47,10 @@
           .then(function (productEntry){
             vm.closeModal();
           }, function (response) {
+            var errors = response.data.errors;
+            angular.forEach(errors, function(error, field) {
+              vm.addAlert(field + " : " + error);
+            });
           });
     };
   }
--- a/app/assets/javascripts/client/inventories/analysisButton.js
+++ b/app/assets/javascripts/client/inventories/analysisButton.js
@@ -1,0 +1,17 @@
+(function() {
+  'use strict';
+
+  angular.module('PDRClient')
+      .directive('analysisButton', analysisButton);
+
+  function analysisButton() {
+    return {
+      restrict: 'E',
+      replace: true,
+      templateUrl: 'client/inventories/analysis_button.html',
+      scope: {},
+      controller: 'AnalysisButtonCtrl',
+      controllerAs: 'analysisButton'
+    }
+  }
+})();
--- a/app/controllers/v1/analyses_controller.rb
+++ b/app/controllers/v1/analyses_controller.rb
@@ -1,0 +1,32 @@
+class V1::AnalysesController < ApplicationController
+  before_action :authenticate_user!
+
+  def create
+    @analysis = inventory.build_analysis(analysis_params)
+    authorize_action_for @analysis
+
+    if @analysis.save
+      render template: 'v1/analyses/show', status: 201
+    else
+      render_error
+    end
+  end
+
+  private
+  def inventory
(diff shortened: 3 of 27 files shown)


bithavoc:
Analyses are being created with `district_id: nil`, is this the intended behavior?

bithavoc:
Attempting to create an analysis in a brand new account without selecting inventory makes the `Create` button seem like no action has been taken, under the hood an API call is made without `inventory_id` so the app gets `No route matches [POST] "/v1/inventories/analyses"`. I believe the button should be disabled until the user is able to select an inventory, if there are no inventories we could come up with something like this:

bithavoc:
The button to create a new analysis looks like this:


Is the styling of this button in the scope of this task?

MakotoTheKnight:
RE https://github.com/MobilityLabs/pdredesign-server/pull/238#issuecomment-211566330:

The scenarios being described here are:

 - A user who has not created an inventory attempts to create an analysis.  This should be explicitly covered by #196.
 - A user who has created an inventory attempts to create an analysis without selecting an inventory to use.  This implies that the inventory selection piece must be a required part of the submission before we allow the user to continue.  Disabling it is an appropriate action. :)

app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
```diff
@@ -0,0 +1,45 @@
...
+
+    vm.analysis = {};
+
+    vm.save = function () {
+      Analysis.create(null, vm.analysis)
+          .$promise
+          .then(function (productEntry){
+            vm.closeModal();
+          }, function (response) {
+          });
```
bithavoc: It doesn't seem like we're taking any action on the error callback, when the `create` call fails, do we want to show the error messages here?

app/assets/javascripts/client/inventories/analysisButton.js
```diff
@@ -0,0 +1,17 @@
+(function() {
+  'use strict';
+
+  angular.module('PDRClient')
+      .directive('analysisButton', analysisButton);
```
bithavoc: I thought we were calling these `<feature>Link` instead.

app/models/analysis.rb
```diff
@@ -0,0 +1,21 @@
...
+#  district_id  :integer
+#  deadline     :datetime
+#  created_at   :datetime
+#  updated_at   :datetime
+#  inventory_id :integer
+#
+
+class Analysis < ActiveRecord::Base
+  include Authority::Abilities
+  self.authorizer_name = 'AnalysisAuthorizer'
```
bithavoc: I believe this authorizer should be automatically picked based on naming convention.

app/assets/javascripts/client/inventories/analysisButton.js
MakotoTheKnight: This is alright for now - we need to clean up the naming conventions around the platform.

app/controllers/v1/analyses_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def render_error
+    @errors = @analysis.errors
+    render 'v1/shared/errors', status: 422
+  end
+
+  def analysis_params
+    # converting from US format because that's what the frontend is sending us
+    params[:deadline] = Date.strptime(params[:deadline], '%m/%d/%Y')
```
bithavoc: I noticed we're using [this same format in inventories](https://github.com/MobilityLabs/pdredesign-server/blob/staging/app/controllers/v1/inventories_controller.rb#L24), I wonder if it's a good time to promote it to a constant.

spec/controllers/v1/analyses_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'spec_helper'
+
+describe V1::AnalysesController do
+  render_views
+
+  let(:inventory) { FactoryGirl.create(:inventory) }
+
+  before :each do
+    request.env['HTTP_ACCEPT'] = 'application/json'
```
bithavoc: Can this be done by using `format: :json` in `post :create, inventory_id: inventory.id, format: :json`?

spec/controllers/v1/analyses_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    it 'creates a record' do
+      sign_in inventory.owner
+
+      post :create,
+           inventory_id: inventory.id,
+           name: "name",
+           district_id: inventory.district.id,
+           deadline: "11/14/2042"
+
+      assert_response 201
```
bithavoc: We've been using the new syntax lately `expect(response).to have_http_status(:created)`.

spec/factories/analyses.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :analysis do
+    name { Faker::Lorem.word }
+    deadline { Faker::Date.between(1.day.from_now, 1.year.from_now) }
+    association :inventory
```
bithavoc: should we include `association :district` in this factory?

app/assets/javascripts/client/inventories/AnalysisModalCtrl.js
juliosantos: I think @MakotoTheKnight's "thumbs up" means he thinks we should. I'll try and get you both on a conversation soon about form validation and errors today. If we're building this from the ground up we should get it right; if we're reusing something I need to know what to reuse.

app/models/analysis.rb
juliosantos: I don't disagree, but this is the pattern that is used throughout the app. I recommend creating a new task to capture that or adding it to the list of codebase improvements @MakotoTheKnight is maintaining.

app/controllers/v1/analyses_controller.rb
juliosantos: Something like this?
```ruby
# app/services/freedom_date.rb
# usage: FreedomDate.parse("11/14/1986")
class FreedomDate
  FORMAT = '%m/%d/%Y'

  def self.parse(date_string)
    Date.strptime(date_string, FORMAT);
  end
end
```
It makes sense to me, maybe with a less allegorical name. What are your thoughts on including it here vs the cleanup list?

spec/controllers/v1/analyses_spec.rb
juliosantos: Absolutely, and I think it would be the better approach. Again, following prevalent patterns. Let me know how you'd like me to proceed.

spec/controllers/v1/analyses_spec.rb
juliosantos: Fixed in 9e65363

spec/factories/analyses.rb
juliosantos: See my other comment re. districts.

juliosantos:
> [@bithavoc] Analyses are being created with district_id: nil, is this the intended behavior?

This is an appendage left over from when a user was required to select a district when creating an analysis, too. I can either remove it or ensure its assignment with an AR callback, if we think it'll help us scope some queries in the near future.

---

> [@bithavoc] Attempting to create an analysis in a brand new account without selecting inventory 

As @MakotoTheKnight said, that's #196

---

> [@MakotoTheKnight] A user who has created an inventory attempts to create an analysis without selecting an inventory to use

This is not a possible journey. The inventory `<select>` is pre-populated with the user's first inventory, and doesn't allow them to deselect it.
That said, f105520 makes it so that the save button is disabled until all the fields are filled out.

juliosantos:
> [@bithavoc] Is the styling of this button in the scope of this task?

I don't know. @MakotoTheKnight what are your thoughts?

MakotoTheKnight:
@juliosantos:  Yes, styling is also required for this task, since that button has an appropriate place to live.   We should be able to borrow the existing style from buttons on that page.

app/controllers/v1/analyses_controller.rb
bithavoc: Let's include this in the cleanup task.

spec/controllers/v1/analyses_spec.rb
bithavoc: let's add it to the cleanup list.

bithavoc:
@juliosantos what I noticed is that the select box is prepopulated with a single inventory, it's automatically selected and still the district_id for the audit is nil when is saved to the database.

juliosantos:
@bithavoc I am aware. Nothing is done at the moment with that field. My question remains:
> I can either remove it or ensure its assignment with an AR callback, if we think it'll help us scope some queries in the near future.

spec/controllers/v1/analyses_spec.rb
juliosantos: @MakotoTheKnight please take heed

app/controllers/v1/analyses_controller.rb
juliosantos: @MakotoTheKnight please take heed

juliosantos:
Handling errors now through the `alert` directive. FWIW, I'm not the biggest fan of this pattern for 3 reasons:

1. It requires us to define `alerts`, `addAlert`, `closeAlert` and the mechanics for populating `alerts` on every controller that uses it.
2. It's not an ideal UX because the field names in the backend might not match the labels given in the frontend.
3. It's not an ideal UX because the errors are all bundled together rather than presented next to the relevant fields.

bithavoc:
LGTM-pending-rebase 🚫 :shipit: until 🕐 ⚾️

juliosantos:
@bithavoc rebased

juliosantos:
@bithavoc all clean now

app/controllers/v1/analyses_controller.rb
MakotoTheKnight: I'll add it to #251.

PR #242, opened 2016-04-18
Closes #200: inventory invites

### LINKED ISSUE #200: User cannot log into account after being invited to an inventory

bithavoc:
Fixes the following scenarios:

> When a **visitor** is trying to signup and there is an **invite to an assessment** in place with that same email address, the user should be redirected to complete the signup process through the assessment invite.

> When a **visitor** is trying to signup and there is an **invite to an inventory** in place with that same email address, the user should be redirected to complete the signup process through the inventory invite.

### DIFF
--- a/app/controllers/v1/user_controller.rb
+++ b/app/controllers/v1/user_controller.rb
@@ -17,8 +17,7 @@ class V1::UserController < ApplicationController
     if user_invitation.present?
       @user.errors.add(:base, "It seems you have already been invited. Please continue your registration here.")
       render json: @user.errors.to_h.merge(
-        invitation_token: user_invitation.token,
-        invitation_type: user_invitation.class.name.underscore
+        invitation_token: user_invitation.token
       ), status: :unprocessable_entity
     elsif @user.save
       send_notification_email
--- a/app/controllers/v1/user_controller.rb
+++ b/app/controllers/v1/user_controller.rb
@@ -7,12 +7,19 @@ class V1::UserController < ApplicationController
   end
 
   def create
-    @user                  = find_by_invite_or_initialize(user_params)
-    @user.role             = params[:role] || :member
-    @user.district_ids     = extract_ids_from_params(:district_ids)
-    @user.organization_ids = extract_ids_from_params(:organization_ids)
+    @user = User.create(user_params.merge(
+      role: params[:role] || :member,
+      district_ids: extract_ids_from_params(:district_ids),
+      organization_ids: extract_ids_from_params(:organization_ids)
+    ))
 
-    if @user.save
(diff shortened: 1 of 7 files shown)


app/controllers/v1/user_controller.rb
```diff
@@ -7,12 +7,19 @@ def show
...
-    @user.organization_ids = extract_ids_from_params(:organization_ids)
-
-    if @user.save
+    @user = User.create(user_params.merge(
+      role: params[:role] || :member,
+      district_ids: extract_ids_from_params(:district_ids),
+      organization_ids: extract_ids_from_params(:organization_ids)
+    ))
+
+    user_invitation = [UserInvitation.find_by(email: user_params[:email]), InventoryInvitation.find_by(email: user_params[:email])].find{|invitation| !invitation.nil?}
```
juliosantos: I prefer:
```ruby
[UserInvitation.find_by(email: user_params[:email]), InventoryInvitation.find_by(email: user_params[:email])].compact.first
```
or
```ruby
[UserInvitation.find_by(email: user_params[:email]), InventoryInvitation.find_by(email: user_params[:email])].find{|invitation| invitation.present?}
```
for readibility.

juliosantos: Or, since I'm at it:
```ruby
[UserInvitation, InventoryInvitation].map{ |i| i.find_by(email: user_params[:email]) }.compact.first
```
:)

juliosantos:
I smoke-tested this and it behaves as expected. I hit when going through the assessment invitation flow (tried to sign up as a new user, was redirected, filled the form, clicked Continue), and then:
```
Started GET "/v1/assessments/237" for ::1 at 2016-04-19 10:46:52 +0100
Processing by V1::AssessmentsController#show as JSON
  Parameters: {"id"=>"237"}
[...]
PG::UndefinedColumn: ERROR:  column messages.assessment_id does not exist
```
Your `schema.rb` does show an `assesstment_id` in the `messages` column, so this is probably an issue with my database state. But I thought I'd bring it up for your consideration anyway.

MakotoTheKnight:
@juliosantos:  #216 adds support for the messages table to be polymorphic, so it no longer relies on `assessment_id`.  I'm thinking that if you're running into this snag, we should be merging #216 in ASAP so that we have that dependency.

bithavoc:
@juliosantos would you give this another look?

juliosantos:
:shipit:

PR #241, opened 2016-04-18
Closes #240: add and require name for data entries

juliosantos:

PR #244, opened 2016-04-19
Closes #209: make product name required

### LINKED ISSUE #209: product_name of a product entry's general inventory question should be required
see #165

juliosantos:


juliosantos:
Closes #209

bithavoc:
I noticed that in [data set name validation](https://github.com/MobilityLabs/pdredesign-server/pull/241) even when the Data Set Name field is not visible(I scrolled down to reach the Save button), the HTML5 error message shows up in the top of the screen and I'm able to see I'm missing the data set name so I scroll up and fill the field. For some reason this is not happening for Product Name, the Save button doesn't process the form but there is no visible hint of what am I missing in the form.

juliosantos:
@bithavoc Well spotted. Should be fixed now. Over to you :)

bithavoc:
LGTM :shipit:

PR #249, opened 2016-04-20
Update annotate gem and automatic annotate on dev database migrations

bithavoc:
What is this?
* `annotate` updated to it's latest version **2.7.0**
* `annotate` will now run automatically upon `rake db:migrate`

I also want us to use the same annotate version across the team and prevent this:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,10 +49,10 @@ GEM
       railties (>= 3.1)
       sprockets (~> 2)
       tilt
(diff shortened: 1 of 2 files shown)


MakotoTheKnight:
We're going to keep this PR around, but not merge it in yet as it is lower priority than work dedicated for the upcoming release.

MakotoTheKnight:
Moving this off the board for now.

PR #247, opened 2016-04-19
Closes #134: inventory dashboard

MakotoTheKnight:
This PR is good to go.  Notes:

 - This follows the style defined in #134 with noted deviations in the listing of learning questions and font styles.
 - There are a lot of styles here that would need to change - notably, the card indicating who has been invited to participate - but I get the impression that this is a more global change than can be contained in the scope of this task.  This also applies to the color of the modal buttons (participant management and sending reminders).
 - The "Modify Schedule" button was not clearly specified and does not function.
 - "Go to Report" does not function due to the page not existing yet.
 - All other buttons on the page should function.

PR #259, opened 2016-04-21
Closes #239:  Mark Inventory as "Up to Date"

### LINKED ISSUE #239: Mark Inventory as "Up to Date"
Users adding products and data sets to an inventory can click the button "Mark Inventory as up to date." The facilitators should be able to see that participants have completed their part of the inventory in the dashboard. 

This is similar to what we have for the RA.

MakotoTheKnight:
This PR is good to go, but depends on #247 to be merged before this can be cleanly merged in.  Notes:

 - This adds functionality to ensure that an inventory can be marked as "complete".
 - Completeness is derived off of the same states that an assessment has (to a degree):
  - User is pending if they have not accepted the invite.
  - User is invited if they have accepted the invite but have not saved their progress (explicitly) when editing the inventory.
  - User is in progress if they have accepted the invite and have explicitly saved their progress while editing the inventory.
  - User is completed with the inventory if they mark it as "complete".
 - A user that has marked an inventory as complete may *not* re-enter the /edit route again.


app/assets/javascripts/client/inventories/Inventory.js
```diff
@@ -24,6 +24,27 @@
       'districtProductEntries': {
         method: 'GET',
         url: UrlService.url('inventories/:inventory_id/district_product_entries')
+      },
+      'markComplete': {
+        method: 'POST',
+        url: UrlService.url('inventories/:inventory_id/mark_complete'),
+        params: {
```
bithavoc: FYI, I believe you already have `inventory_id` in the URL, so in the rails controller `inventory_id` will still be in `params[:inventory_id]`. Also, the restful way of doing this would be PATCHing the inventory with the right attribute to modify it's state, so `PATCH /inventories/:inventory_id, { response.submitted_at: Time.now}` would have been enough.

We can add this to the cleanup list.

MakotoTheKnight: That's not a difficult change to make.  It can be done quickly enough.

MakotoTheKnight:
@bithavoc:  Mind giving this another look?

bithavoc:
@MakotoTheKnight I'm not finding any errors during the process(except for the missing `inventories_report` stat to be filled in by https://github.com/MobilityLabs/pdredesign-server/pull/260), however, I couldn't get an inventory to it's completed state.

app/assets/javascripts/client/inventories/Inventory.js
bithavoc: sorry for the lack of details in my original comment:
I don't think we need a new endpoint for this as PATCH'ing the inventory resource would be enough. We just need to update the inventory with that given atribute to modify it's state and therefore perform the desired action.

So instead of `POST /inventories/:inventory_id/mark_complete` (RPC style) we could be doing `PATCH /inventories/:inventory_id` with body `submitted_at=12-04-2015`. If you don't trust the client to set the date, you can still receive `submitted_at=true` and override it internally with `Time.now` and still be perfectly RESTful.

Same with `PUT /inventories/:inventory_id/save_response`(RPC style), it should have been `POST /inventories/:inventory_id/responses`, and since it's idempotent as there should be only one response per user, it could have been `POST /inventories/:inventory_id/responses/:user_id`. Even ``PATCH /inventories/:inventory_id` with body `replied=true` could do the trick in this case. 

Again, these are just suggestions and I would consider them an improvement, it works as it is.

bithavoc:
LGTM

PR #262, opened 2016-04-22
195 analysis create page

### LINKED ISSUE #195: Analysis Create Page
We should be able to reuse a lot of what we put together for the Inventory Create Page. We will need to have a listing of inventories for users to select from in Step 1. Participants and facilitators from that inventory should be brought over from the Inventory. Facilitators can add/manage additional facilitators, but not participants.

juliosantos:
Please do smoke-test this intensively. I believe I lifted all the behaviour I saw with inventories, but something might have slipped through the cracks.

Closes #195

### DIFF
--- a/app/assets/javascripts/client/common_create/CreateService.js
+++ b/app/assets/javascripts/client/common_create/CreateService.js
@@ -197,7 +197,7 @@
       } else if (service.context === 'inventory') {
         return InventoryParticipant.query({inventory_id: service.extractId()});
       } else if (service.context === 'analysis') {
-        return AnalysisParticipant.query({inventory_id: service.extractId()});
+        return AnalysisParticipant.query({inventory_id: service.extractId(), analysis_id: $stateParams.id});
       }
     };
 
@@ -215,6 +215,7 @@
       } else if (service.context === 'analysis') {
         return AnalysisParticipant.delete({
           inventory_id: $stateParams.inventory_id,
+          analysis_id: $stateParams.id,
           id: participant.participant_id
         }).$promise;
       }
@@ -228,7 +229,7 @@
         return InventoryParticipant.query({inventory_id: service.extractId()})
             .$promise;
       } else if (service.context === 'analysis') {
-        return AnalysisParticipant.query({inventory_id: service.extractId()})
+        return AnalysisParticipant.query({inventory_id: service.extractId(), analysis_id: $stateParams.id})
             .$promise;
       }
     };
@@ -241,7 +242,7 @@
         return InventoryParticipant.all({inventory_id: service.extractId()})
             .$promise;
       } else if (service.context === 'analysis') {
-        return AnalysisParticipant.all({inventory_id: service.extractId()})
+        return AnalysisParticipant.all({inventory_id: service.extractId(), analysis_id: $stateParams.id})
             .$promise;
       }
     };
@@ -253,7 +254,8 @@
         }, {user_id: user.id}).$promise;
       } else if (service.context === 'analysis') {
         return AnalysisParticipant.create({
-          inventory_id: $stateParams.inventory_id
+          inventory_id: $stateParams.inventory_id,
+          analysis_id: $stateParams.id
         }, {user_id: user.id}).$promise;
       }
     };
--- a/app/assets/javascripts/client/common_create/CreateService.js
+++ b/app/assets/javascripts/client/common_create/CreateService.js
@@ -161,7 +161,7 @@
       }
     };
 
-    service.saveAnalysis = function(analysis) {
+    service.saveAnalysis = function(analysis, assign) {
       if (analysis.name === '') {
         service.emitError('Analysis needs a name!');
         return;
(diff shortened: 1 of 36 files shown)


bithavoc:
I'm able to press the button to create a new analysis but the modal doesn't show me anything, is this part of the scope for this PR?

bithavoc:
pressing the button `Send Invitations` doesn't send out the invitation emails so I'm unable to test the send invite/accept flow.

bithavoc:
Spotted some differences between the design and the implementation here, so this is the [design attached here](https://github.com/MobilityLabs/pdredesign-server/issues/195):



This is the newly introduced assign page for the inventory at `inventories/1/analysis/assign`:

screen shot 2016-04-22 at 3 41 41 pm

I'm not sure if the user is supposed to change the district as it should be using the district already set in the inventory.

juliosantos:
@bithavoc: Thanks for the review. See below for comments.

> I'm able to press the button to create a new analysis but the modal doesn't show me anything, is this part of the scope for this PR?

It's not part of the scope of this PR, but it does reveal what looks like it might have been a merge error, or a failed conflict resolution. I pushed another commit up with the fix. Thanks for pointing this out.

> pressing the button Send Invitations doesn't send out the invitation emails so I'm unable to test the send invite/accept flow

The "accept" flow isn't to my knowledge in scope. But the emails should have been sent. I think I was confused with the different kinds of emails and didn't notice sidekiq erroring out for one of them. I pushed up a fix, please test this again.

> Spotted some differences between the design and the implementation here

The implementation is consistent with assessments and inventories. I suppose at some point we'll want to change the CSS for all those i one fell swoop.

> I'm not sure if the user is supposed to change the district as it should be using the district already set in the inventory.

Well spotted. I agree, and it's a similar discussion to what we had with the new analysis modal. That form is a generic `tool_details` form, which doesn't allow for hiding the district field at the moment. We could extend it, but I'd rather talk to you guys before wrt the how.

bithavoc:
We're going to need a new column `assigned_at` in the `Analysis` model just like we have one for [ assessments](https://github.com/MobilityLabs/pdredesign-server/blob/staging/db/schema.rb#L57).

MakotoTheKnight:
Some things I noticed in the review:

 - Learning questions are bound to the inventory and not the individual analysis.  This will become straightforward to address when we specify the analysis we're creating (once the relationship becomes a one-to-many).
 - There's no indication as to which inventory is selected.  This should be available to the user in Step 1 (as Johan has pointed out).  We will need to augment the `tool-details` directive to include a context and incorporate an `ng-if` on the view to hide the inventory selector from showing up if we're not in the context of an analysis.

 That said, we're also going to want to figure out what to do with that; do we ever want that to change?  It doesn't make immediate sense to me to allow that to be modified.

 - Participants from the original inventory are not participants of the analysis.
 - We can invite non-facilitators to the analysis.  Depending on the amount of work this may take, we can break this off into a separate task to be worked.

Do what you feel you need to do to accommodate the use cases we have for analyses (including disabling the district and extending the directive(s) you need to).

juliosantos:
> [@bithavoc] We're going to need a new column `assigned_at` in the `Analysis model`

Done. I lifted the relevant behaviour from `Assignment`.

> [@MakotoTheKnight] Learning questions are bound to the inventory and not the individual analysis

Not any more :)

> [@MakotoTheKnight] There's no indication as to which inventory is selected

I changed the `toolDetails` directive so that you can disable the district dropdown.

> [@MakotoTheKnight] Participants from the original inventory are not participants of the analysis.
> [@MakotoTheKnight] We can invite non-facilitators to the analysis

I don't understand these. Will bring it up at standup.

MakotoTheKnight:
Thanks for the changes @juliosantos.  Feedback:

 - We still don't have an indication as to which inventory this analysis belongs to.  Original design documents display some sort of drop-down menu that contains our current inventory.  Pending conversations with @HeyCat this may block the conclusion of this task.
 - We've discussed inviting facilitators and what the status of that was, but for clarity, it's fine if we can invite both for now since we don't anticipate the typical facilitator to invite participants to the analysis.
 - We've got a similar issue with accounts being created in the invite flow as we did with inventories a while back; if we create an account and invite them with those credentials, that account cannot be logged in to.
 - Text-based emails show HTML tags.  This may be a holdover from inventories in which it does a similar thing, so we will create a task to capture this.

To solve the account creation issue, I'd rather move this out into a separate body of work since that is effort on its own.  @bithavoc has done work to that effect beforehand and may be able to point you in the right direction.

Check in with our design team to see if what we've got is good to roll with as is or if we do want to add in the inventory name somewhere on the screen, even if we can't change it.

juliosantos:
@MakotoTheKnight I don't get what you mean with this:

> We still don't have an indication as to which inventory this analysis belongs to

Look:

screen shot 2016-04-29 at 12 24 28

As for your other comments, let's pick this up at standup.

MakotoTheKnight:
LGTM.  After you rebase, :shipit:!

PR #269, opened 2016-04-25
Closes #192: Add links to the inventory list view

### LINKED ISSUE #192: Add inventory links in the list of inventories
The list of inventories and assessments is implemented between `common_list` and `entry_view`, assessment already works so for inventories we'd need to add a condition in `entry_view.html` similar to what we already do for assessments.

Example:

```
<div class="links" ng-if="isInventory === 'true'">
        <div class="repeat" ng-repeat="link in entry.orderLinks(entity.links)">
          <inventory-links
              data-title="{{link.title}}"
              data-active="{{link.active}}"
              data-type="{{link.type}}"
              data-id="{{entity.id}}"
              data-consensus-id="{{entity.consensus.id}}"
              data-role="{{role}}">
          </inventory-links>
  </div>
</div>
```

MakotoTheKnight:

 - Links pointing to different facets of an inventory have been created.  Note that the analysis link may not yet work due to that work depending on another PR.  We direct ourselves to the route which would help us create a new analysis; if we should use a modal instead, we can wire that in when we integrate this branch together.
 - All facets should be checked:  facilitator and participant, with and without access to an inventory.
 - This PR also corrects the inventory index query; we should pull back all inventories that match the current user's district(s), even if they don't have access.


app/assets/javascripts/client/directives/permissionsInfo.js
```diff
@@ -1,21 +1,25 @@
...
+              placement: scope.placement,
+              trigger: 'hover',
+              content: '<i class="fa fa-bullhorn"></i><span>Facilitator</span><p>Facilitators share responsibility with the district facilitator to contact participants and view their individual responses, facilitate the consensus meeting, and view the final consensus report.</p><i class="fa fa-pencil-square-o"></i><span>Participant</span><p>Participants responds to the individual readiness assessment survey and take part of the consensus meeting. They can view the final consensus report</p>'
+
+            });
+          },
+          templateUrl: 'client/views/directives/help_info_popover.html'
+        }
+      })
+})();
```
juliosantos: Not really crucial, but we've been using named function declarations for the directive and its link.

app/controllers/v1/inventory_access_requests_controller.rb
```diff
@@ -8,7 +8,9 @@ def index
 
   def create
     permission = Inventories::Permission.new(inventory: inventory, user: current_user)
-    @request = permission.request_access(role: create_request_params[:role])
+    # Note:  UI passes a list but the UI will never send back more than one element.
+    # UI has not been changed to preserve compatibility with existing code which assumes the same.
+    @request = permission.request_access(role: create_request_params[:roles].try(:[], 0))
```
juliosantos: Isn't `try` cool? :) I mean, I'm sure Sandi Metz would be all like "omg nothing is something and this is a hidden conditional!!!111shift", and she does have a point, but it's so damn convenient.

lib/link/inventory.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    private
+    def target
+      return Link::InventoryFacilitator if facilitator?(user: user)
+      return Link::InventoryParticipant if participant?(user: user)
+      Link::Partner
+    end
+  end
+end
```
juliosantos: Functionality in `/lib` is conventionally reserved for cross-application library code, akin to gems. This should really be in `app/services` or similar.
I know this is the pattern we've been following, but thought I'd mention this anyway.

juliosantos:
:shipit:

app/assets/javascripts/client/directives/permissionsInfo.js
MakotoTheKnight: Fair comment; this is an older directive that we're going to want to rework.  I didn't do it here since, if I were going to go that full route I'd want to move it into its appropriate spot at the same time.

lib/link/inventory.rb
MakotoTheKnight: We should capture this in a cleanup task so that we can do this.  Yes, this should ideally be a service, and we should move all of them over at once. :smile:

PR #272, opened 2016-04-26
Closes #265: Make Rubrics polymorphic

### LINKED ISSUE #265: Analysis - Make rubrics polymorphic
To support allowing analyses create a consensus, make rubrics polymorphic so that we can hook into this directly.

- [x] Migrate to polymorphic type "tool"
- [x] Ensure there are indexes based on type
- [x] Migrate all existing rubrics to use "Assessment"

MakotoTheKnight:

 - We had originally believed that Rubric needs to be polymorphic to accept any tool ID and tool type, but this turned out not to be the case.  The only change in regards to the model that we needed to make was to add a `tool_type` column which is indexed, and ensure that anything looking for an assessment rubric pulled one with a `tool_type` of `'Assessment'`.
 - This removes some of the crufty bits associated with Rubric, and removed an unused relationship between a Rubric and a User.
 - This directly depends on #262, so I'd prefer to have that reviewed and merged into staging before this one.


juliosantos:
LGTM

PR #278, opened 2016-04-28
Update forms to submit properly and remove extra divs

### LINKED ISSUE #133: Reset password boxes don't seem to align with new style
This is what it looks like in staging:

image

This is what it looks like on production:

image

seanperkins:
Closes #133 

Major change here is:
Using ngSubmit on form instead of ngClick on a submit button. It has the added benefit of HTML5 Form Validations.

You will also notice I got rid of some divs and put the form class on the form element per bootstrap expectations. The page is now responsive the way one would expect.

### DIFF
--- a/app/assets/javascripts/client/views/reset_password/request_reset.html.erb
+++ b/app/assets/javascripts/client/views/reset_password/request_reset.html.erb
@@ -1,64 +1,55 @@
 <div class="login signup">
   <div class="row secondary">
-    <div class="container">
-      <div class="col-md-4">
-        <h1>Forgot your password?</h1>
-        <h3 class="subheading" >We'll email you instructions to reset your password.</h3>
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/client/views/reset_password/request_reset.html.erb
```diff
@@ -1,64 +1,55 @@
...
+          <input
+            required
+            ng-model="email"
+            class="form-control"
+            placeholder="Email Address*"
+            type="email">
+          <input
+            ng-disabled="resetForm.$invalid"
+            type="submit"
+            class="btn btn-devise btn-danger pull-right"
```
MakotoTheKnight: I wonder what this `btn-devise` class represents.  There's no evidence if of it in the existing codebase.

MakotoTheKnight:
LGTM :shipit:

PR #289, opened 2016-04-29
199 analysis dashboard page

### LINKED ISSUE #199: Analysis Dashboard Page
This is very similar to the Inventory Dashboard.

bithavoc:
Closes #199


MakotoTheKnight:
Would you mind rebasing this branch against staging?

MakotoTheKnight:
This is something we missed in an earlier PR, but the [AnalysisAuthorizer](https://github.com/MobilityLabs/pdredesign-server/blob/staging/app/authorizers/analysis_authorizer.rb) makes some strong assumptions about where permission comes from.  We effectively copy over the user and create its own `AnalysisMember` entities, and we should be looking at that instead of the inventory.  This doesn't directly impact your PR in that it doesn't concern code that you've authored, but it _is_ a break in program flow since it means that you cannot ever see an analysis unless you're a facilitator.

app/assets/javascripts/client/inventories/analysis_dashboard.html
```diff
@@ -0,0 +1,70 @@
...
+        </div>
+      </div>
+      <hr class="dashboard-hr" />
+      <div class="col-md-6 col-sm-6" ng-repeat="user in analysisDashboard.participants">
+        <participant user="user"></participant>
+      </div>
+    </div>
+    <div class="row">
+      <div class="col-lg-12">
+        <add-inventory-users-link send-invite="true"></add-inventory-users-link>
```
MakotoTheKnight: This will not work for an analysis.  This manages the inventory's permissions.  We have modals to use for this flow instead; refer to [`common_create/participants.html`](https://github.com/MobilityLabs/pdredesign-server/blob/199-analysis-dashboard-page/app/assets/javascripts/client/common_create/participants.html#L44).

app/assets/javascripts/client/inventories/analysis_dashboard.html
```diff
@@ -0,0 +1,70 @@
...
+  <div class="row row-bg row-header">
+    <div class="col-md-12">
+      <h1 class="greeting">{{analysisDashboard.inventory.name}}</h1>
+      <h4 class="byline">organized by: <a href="#">{{analysisDashboard.inventory.facilitator.full_name}} </a>
+        on {{analysisDashboard.inventory.created_at | amDateFormat:'MMMM Do, YYYY'}}
+      </h4>
+    </div>
+  </div>
+  <div class="row">
+    <div id="participants" ng-if="analysisDashboard.currentAnalysis.is_facilitator">
```
MakotoTheKnight: This facilitator check will _hide_ the participants list as opposed to showing it.  I believe an earlier version of the inventory dashboard did something similar but it has [since been corrected](https://github.com/MobilityLabs/pdredesign-server/blob/199-analysis-dashboard-page/app/assets/javascripts/client/inventories/dashboard.html#L23).  The if-check needs to be moved to the div containing the invitation control pieces.

app/assets/javascripts/client/inventories/analysis_dashboard.html
```diff
@@ -0,0 +1,70 @@
...
+    <div id="messages" ng-if="analysisDashboard.inventory.messages">
+      <div class="col-md-12 header">
+        <h2 ng-pluralize count="analysisDashboard.inventory.messages.length"
+                          when="{'0': 'No Sent Messages',
+                                 '1': '{} Sent Message',
+                                 'other': '{} Sent Messages'}">
+
+        </h2>
+      </div>
+      <div class="col-md-12" ng-repeat="message in analysisDashboard.inventory.messages">
```
MakotoTheKnight: This shouldn't be pulling the inventory's messages.  An analysis has its own messages that it can pull in.

bithavoc:
@MakotoTheKnight check this again pls

bithavoc:
I guess this should say "Individual Analysis Reminder" instead, will fix:

MakotoTheKnight:
The emails seem to be missing some information:

screenshot_2016-05-10_08-55-16

This may fall slightly out of scope but let's confirm with @candyj0311 to see what the copy on this needs to be.

MakotoTheKnight:
When travelling between inventories and analyses, the flow of the buttons seems to be incomplete.

For example, when going from an inventory to the analysis, the button "Go to analysis" does not work with this error:

```Could not resolve 'analysis_dashboard' from state 'inventories_report'```

When going from an analysis back to the inventory it works just fine.

File: app/assets/javascripts/client/inventories/MessageElementCtrl.js [hunk 4]
```diff
@@ -6,12 +6,13 @@
 
   function MessageElementCtrl() {
     var vm = this;
-    vm.messageTitle = function(category) {
+    vm.messageTitle = function(category, kind) {
+      var subjectKind = kind || 'Inventory';
```
MakotoTheKnight: It feels like you should pass this in from the directive instead.  The functionality would still be the same, but you'd be getting the `subjectKind` from `$scope`.

bithavoc: should I change the way we're passing the category in too? we're doing the same thing for the category, it's actually `message.category` that I can definitely get via `$scope.message.category`. it's no pain.

bithavoc:
can you try this functionality one more time?

bithavoc:
@MakotoTheKnight can you try this functionality one more time?

bithavoc:
This PR is good to go with a few additions:
* Added learning questions to the sidebar (it wasn't in the designs initially)
* Tweaked Reminder template, not it looks more like **Analysis**
* Modify calendar(formerly known as **Modify schedule**) was added to the sidebar. We may need to add **GMT midnight** correction to the deadline at some point, we're sending to the server **31-05-2016** but is being displayed in the analysis schedule summary as **30-05-2016** which would be `31-05-2016:00:00:00` - `5 hours` adjusted locally. We have the same situation in inventories as well.
* Added kind to `message-element` so it no longer reports the reminder entries titled as `Regular Inventory Reminder` but `Regular Analysis Reminder` instead.
* Further behavior of the *Go to Analysis* button are being discussed in https://github.com/MobilityLabs/pdredesign-server/issues/308

MakotoTheKnight:
- Emails sent off in an invite context for analyses aren't sent immediately.  Both inventories and assessments have this functionality.
 - The title, organizer and creation time of the analysis are direct properties of the inventory.
 - The participants' status is not present on the screen.

A GIF of the last two issues can be found below:

anim

bithavoc:
@MakotoTheKnight Also found that in inventory dashboard(app/assets/javascripts/client/inventories/dashboard.html) we had attributes set too but they don't seem to take any effect:
`<add-inventory-users-link send-invite="true"></add-inventory-users-link>`
`<manage-inventory-permissions-link send-invite="true"></manage-inventory-permissions-link>`

These directive are also expecting `inventoryId`, same with `addAnalysisUsers` seems to be expecting `analysisId` but it's not being set nor used anyway, check `app/assets/javascripts/client/inventories/addAnalysisUsers.js`

 ```
scope: {
  inventoryId: '=analysisId'`
}
```

The only one that works is:
`invite-inventory-user-link send-invite="true`

That was the one that I ported over by adding the `send-invite` flag and its sending the email but it's still missing the flow to accept invitations for analysis as in `http://localhost:3000/#/inventories/1/analyses/1/invitations/243b82947c` which I think it should be part of a separate task.

MakotoTheKnight:
We're content to merge this work and deploy it into staging.

Issue #268, opened 2016-04-25
Analysis Consensus - Frontend (fully wired)


MakotoTheKnight:
Fully wire in the consensus frontend with the backend.

Depends on #266

MakotoTheKnight:
We have a pop-up modal associated with the readiness assessment that warns a user if they're finishing the consensus prematurely.  Do we wish to incorporate that here as well?

candyj0311:
@MakotoTheKnight Yes, we should.

juliosantos:
@candyj0311 Please note that this piece will be reusing components currently in place for an assessment consensus. When QAing, plese ensure no regressions occurred with those.

PR #296, opened 2016-05-03
Closes #266:  Analysis Consensus Controller

### LINKED ISSUE #266: Analysis Consensus - Controller
We require a controller to handle the same functionality as the existing ConsensusController, which is geared toward assessments.

This also includes adding the questions to the database (found  at https://docs.google.com/spreadsheets/d/1ladm2pK-bSp9VVSA8mra8VsGPg1p6GxG2klrEY8ROFE/edit#gid=0).  Note that "Category" translates to "Question" in the backend, and the "Category Definition" should actually be the body of the question.

MakotoTheKnight:

 - This effectively replicates the flow for creating a consensus and creating an analysis.  The two of these controllers are similar, but they do not share any code.
 - There was a bit more work done with some of the views for the `AnalysisConsensusController` specifically.  This is due to there being an inheritance hierarchy between the `ConsensusController` and `AnalysisController` that I didn't see much value in porting over.
 - Much of the existing code was able to be reused with this work.  The main bits that have changed are the portions which add the extended functionality for data and inventory entries.
 - This cannot be smoke tested through any UI at this time (#268 will utilize this), but it can be tested through the use of a REST client.
 - This makes the assumption that the `AnalysisResponse` is posted to _first_, similar to how responses and consensus are created with the Readiness Assessment.


app/authorizers/response_authorizer.rb
```diff
@@ -10,13 +10,13 @@ def updatable_by?(user)
 
   def readable_by?(user)
     return false unless resource.responder
-    return true  if resource.responder_type == 'Assessment'
+    return true if %w(Assessment Analysis).include? resource.responder_type
```
juliosantos: This is nitpicking, but:
```ruby
return true if resource.responder_type.in? %w(Assessment Analysis)
```
might be more readable.

app/controllers/v1/analysis_responses_controller.rb
```diff
@@ -0,0 +1,37 @@
+class V1::AnalysisResponsesController < ApplicationController
+  before_action :authenticate_user!
+
+  def create
+    @response = Response.find_or_create_by(responder: analysis)
```
juliosantos: If you're calling `.save` below, using `Response.find_or_initialize_by` could save you a query.

app/controllers/v1/analysis_responses_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    authorize_action_for @response
+
+    @response.rubric = analysis.rubric
+    if @response.save
+      create_empty_scores(@response)
+      render nothing: true, status: :created
+    else
+      render nothing: true, status: :bad_request
```
juliosantos: Should we be rendering errors?

MakotoTheKnight: This is a holdover from existing behavior.  Presently the `ResposnesController` will render nothing on an error.  The other caveat to this is, considering that creating a response can't realistically fail - no validations are run on the model - there's no need in my mind to render errors that won't occur.  The intent here is that if this does change, our tests should cover this.

app/controllers/v1/scores_controller.rb
```diff
@@ -1,52 +1,69 @@
...
     end
   end
 
-  private
-  def find_or_initialize
-    Score.find_or_initialize_by(
-      response_id: params[:response_id],
-      question_id: params[:question_id])
+  def fetch_score(question:, response:)
+    Score.includes(:supporting_inventory_response).where(question: question, response: response).first
```
juliosantos: Nitpicking again: `.find_by` instead of `.where` saves you the `.first`

spec/authorizers/response_authorizer_spec.rb
```diff
@@ -0,0 +1,486 @@
+require 'spec_helper'
+
+describe ResponseAuthorizer do
+  describe '#creatable_by?' do
+    let(:user) {
+      create(:user)
```
juliosantos: Love that we're dropping the `FactoryGirl.` prefix :)

juliosantos:
0. None of my comments is a showstopper.
0. This will be a lot easier to test and reason about if I can base my WIP frontend piece on this.

So: LGTM

PR #305, opened 2016-05-09
Closes #301: Cleanup recovery flow

### LINKED ISSUE #301: Clean up recovery flow with additional check
Speak to @MakotoTheKnight for more details.

MakotoTheKnight:
This PR is good to go.  For details on how to test it, please see @MakotoTheKnight.

### DIFF
--- a/app/controllers/v1/user_controller.rb
+++ b/app/controllers/v1/user_controller.rb
@@ -3,25 +3,28 @@ class V1::UserController < ApplicationController
   before_action :downcase_email
 
   def show
-    render partial: 'v1/shared/user', locals: { user: current_user }
+    render partial: 'v1/shared/user', locals: {user: current_user}
   end
 
   def create
     @user = User.create(user_params.merge(
-      role: params[:role] || :member,
-      district_ids: extract_ids_from_params(:district_ids),
-      organization_ids: extract_ids_from_params(:organization_ids)
+        role: params[:role] || :member,
+        district_ids: extract_ids_from_params(:district_ids),
+        organization_ids: extract_ids_from_params(:organization_ids)
     ))
 
-    user_invitation = [UserInvitation, InventoryInvitation].map{|i| i.find_by(email: user_params[:email])}.compact.first
+    user_invitation = [UserInvitation, InventoryInvitation].map { |i|
+      i.find_by(email: user_params[:email])
+    }.compact.first
+    invite_user = user_invitation.try(:user)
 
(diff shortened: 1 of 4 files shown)


spec/controllers/v1/user_controller_spec.rb
```diff
@@ -184,42 +188,122 @@ def post_create_user(options = {})
       kim = User.find_by(email: 'kim@gov.nk')
       expect(kim.organizations.count).to eq(2)
     end
-    
+
     it 'can find and update an existing user' do
```
juliosantos: Shouldn't this be "can't"?

app/controllers/v1/user_controller.rb
```diff
@@ -3,27 +3,34 @@ class V1::UserController < ApplicationController
...
 
-    user_invitation = [UserInvitation, InventoryInvitation].map{|i| i.find_by(email: user_params[:email])}.compact.first
-    if user_invitation.present?
-      @user.errors.add(:base, "It seems you have already been invited. Please continue your registration here.")
+    user_invitation = [UserInvitation, InventoryInvitation].map { |i|
+      i.find_by(email: user_params[:email])
+    }.compact.first
+    invite_user = user_invitation.try(:user)
+
+    if invite_user && invite_user.sign_in_count > 0
```
juliosantos: After a user redeems an invitation, `invite_user` should be `nil`. I don't see how this branch is ever reached.

juliosantos:
It could be that I'm misunderstanding the scope, but this still allows impersonating someone that hasn't redeemed their invitation yet.

MakotoTheKnight:
@juliosantos:  You're right.  This PR wasn't intended to address that aspect of it.  This PR was intended to guard against a user with an existing invitation in the system who had previously signed in, but did not have their invitation deleted.  Effectively, this doesn't work for brand new accounts, but it does work for the legacy accounts in our system, of which we've got ~192.

I'd almost prefer to investigate how it's possible that accounts got into that state prior to the day we implemented the fix, but that's not entirely germane to guarding against this possibility.

app/controllers/v1/user_controller.rb
MakotoTheKnight: I've added some notes and did a bit of looking around.  We have the existing state in our code base in that there are still some invitations lurking about, and there are accounts with a lot of sessions logged (some over 100).

MakotoTheKnight:
I will be tentatively merging this work in. There will be instructions on how to test it in our Slack channel.

PR #306, opened 2016-05-10
Closes #267: Analysis consensus frontend partial

### LINKED ISSUE #267: Analysis Consensus - Frontend (partially wired)
Create the layout for the consensus page.

The consensus pieces should not be wired in at this time, but the Learning Questions are eligible to be.

juliosantos:
Since there is no analysis list from where to generate/view a consensus, in order to test this,

0. Create an analysis on an inventory
0. Create the consensus for that analysis by navigating to `inventories/:inventory_id/analyses/:analysis_id/consensus`
0. The consensus should be created and you should be redirected to `inventories/:inventory_id/analyses/:analysis_id/consensus/:consensus_id`


bithavoc:
**Minor:** When searching for either data or product entries the search term only seems to be matching the beginning of the entry titles.

**Example full listing**


**Example product not matching with name:**




@candyj0311 thoughts?

bithavoc:
Upon saving a response, the sidebar should be updated with the response count:



Only after _refreshing the page_ I can see it updated:



**Additionally:** it seems like we're using "Save **Response**" in the buttons but the sidebar reports them as "Saved **Answers**", shouldn't we be using a single term to describe the same concept?

bithavoc:
When about to submit concensus, should **Save for Later** and **Submit concensus** both redirect to the home page upon saving? I'd think it should at least redirect to either the list of analyses or the inventory page:


 @candyj0311, thoughts?

bithavoc:
There seems to be a regression in readiness assessment:


**Staging** branch:

juliosantos:
> [@bithavoc] There seems to be a regression in readiness assessment
How did you get to that route?

bithavoc:
@juliosantos 
> How did you get to that route?

Try `http://localhost:7130/#/assessments/45/responses/102`

juliosantos:
> [@bithavoc] When searching for either data or product entries the search term only seems to be matching the beginning of the entry titles

Not quite: it's case-sensitive. I still don't like it though and will change it to case-insensitive.

bithavoc:
@juliosantos 
> Not quite: it's case-sensitive. I still don't like it though and will change it to case-insensitive.

Point was that it shouldn't just match case insensitive words at the beginning of the description but all the words in the description.

juliosantos:
@bithavoc: I'm not sure what you mean. It does match any word, case-sensitively. I will change this by making it match any word, case-insensitively. Is this not what you mean?

juliosantos:
> [@bithavoc] the sidebar should be updated with the response count
Well spotted. Fix on the way.

juliosantos:
@bithavoc The regression you spotted should now be fixed. First time I even saw that page - ouch!

bithavoc:
Functionality-wise this PR seems to be working properly now.

app/assets/javascripts/client/consensus/ConsensusService.js
```diff
@@ -0,0 +1,153 @@
...
+      return service;
+    };
+
+    service.extractId = function() {
+      return $stateParams.assessment_id || $stateParams.analysis_id;
+    };
+
+    service.redirectToIndex = function () {
+      if (service.context === "assessment") {
+        $location.path("/assessments");
```
bithavoc: I wonder if we should be using `$state.transitionTo` here instead, although if it fetches the same results then I don't a reason to change it now.

app/assets/javascripts/client/consensus/ConsensusService.js
```diff
@@ -0,0 +1,153 @@
...
+
+  function ConsensusService($stateParams, $location, $q, Consensus, AnalysisConsensus, AnalysisResponse, Score, AnalysisConsensusScore, ProductEntry, DataEntry) {
+    var service = this;
+
+    service.setContext = function(context) {
+      service.context = context;
+      return service;
+    };
+
+    service.extractId = function() {
```
bithavoc: I've seen this in a couple of places now, the name `extractId` doesn't really tell me what `ID` is being extracted, is it the `currentId` or the `parentId`?

app/assets/javascripts/client/consensus/ConsensusService.js
```diff
@@ -0,0 +1,153 @@
...
+          inventory_id: $stateParams.inventory_id,
+          analysis_id: service.extractId()
+        }, {}).$promise
+        .then(function () {
+          return AnalysisConsensus.create({
+            inventory_id: $stateParams.inventory_id,
+            analysis_id: service.extractId()
+          }, {}).$promise;
+        });
+      }
```
bithavoc: I consider this sort of "blind" `if`'s a bit dangerous as it will probably increase the effort required to introduce a new context in addition to `assessment` and `analysis`, specially since the method won't have any effect when a new context is used and therefore fail silently since the callers are not really needing the returned promise. Hard to debug.

app/assets/javascripts/client/views/consensus/header.html
```diff
@@ -16,15 +16,10 @@
     <div class="media">
       <div class="media-body">
           <h2>Instructions</h2>
-          <p class='instructions-content'>
-            The Readiness Assessment consists of questions across the 8 categories of the <readiness-assessment-modal title="PD System Map"></readiness-assessment-modal>. As the assessment facilitator, use this page to guide and document the conversation during the in-person consensus meeting.
-          </p>
-          <p class='instructions-content'>
-            For each question, discuss the individual responses as a group to collectively decide a consensus score, including notes for reference. You may sort or skip questions to support a targeted discussion. Save each completed response as you move through the consensus meeting.
-          </p>
+          <p class='instructions-content' ng-repeat='paragraph in instructionParagraphs()' bind-html-compile='paragraph'></p>
```
bithavoc: this approach is interesting, but I don't see the diference between this and a template that renders conditionally based on context. The templates seems more easy to understand from my point of view.

app/views/v1/analysis_consensus/show.json.jbuilder
```diff
@@ -2,12 +2,14 @@ json.id @response.id
...
 json.is_completed @response.submitted_at.present?
 
+number = 0
 json.categories @categories do |category|
   json.id category.id
   json.name category.name
   json.axis_id category.axis_id
 
   json.questions category.rubric_questions(@rubric) do |question|
+    json.number number += 1
```
bithavoc: shouldn't this be `json.number = question.order + 1`?

app/assets/javascripts/client/inventories/analysis_consensus_answer.html
```diff
@@ -0,0 +1,58 @@
...
+  </div>
+  <div class="score">
+    <div ng-repeat="answer in question.answers">
+      <span class="score-title">{{analysisConsensusAnswer.answerTitle(answer.value)}}</span>
+      <label class="score-checkbox scored-{{answer.value}}">
+        <fake-radio proxy-ng-model="question.score.value" proxy-ng-value="answer.value"></fake-radio>
+      </label>
+    </div>
+    <!-- Temporary hack for hiding "skip". -->
+    <div ng-if="0 == 1">
```
bithavoc: should be using `===` here?

app/assets/javascripts/client/inventories/inventory_routes.js
```diff
@@ -163,6 +163,40 @@
...
+          return Analysis.get({inventory_id: $stateParams.inventory_id, id: $stateParams.analysis_id}).$promise;
+        }],
+        consensus: ['$stateParams', 'AnalysisConsensus', function($stateParams, AnalysisConsensus) {
+          return AnalysisConsensus.get({inventory_id: $stateParams.inventory_id, analysis_id: $stateParams.analysis_id, id: $stateParams.id}).$promise;
+        }]
+      },
+      views: {
+        '': {
+          controller: 'ConsensusShowCtrl',
+          //controllerAs: 'consensusShow',
```
bithavoc: I guess we don't need this then.

bithavoc:
@juliosantos left a few comments, none of them are critical though. We may need a rebase against `staging` before we proceed.

app/assets/javascripts/client/consensus/ConsensusService.js
juliosantos: I don't oppose it but don't see the point either.

app/assets/javascripts/client/consensus/ConsensusService.js
juliosantos: Neither: it's extracting the *relevant* ID. This `ConsensusService`, like others before it, is designed to abstract away the particularities of the entities it works with (in this particular case, assessments and analyses).
Note that there's only one of `$stateParams.assessment_id` or `$stateParams.analysis_id` at any given route, which is why this works (i.e. `$stateParams.assessment_id && $stateParams.analysis_id` is always `undefined`.

app/assets/javascripts/client/consensus/ConsensusService.js
juliosantos: I agree with you but am a lot less concerned. Should you need to wire in a consensus for a new type of tool, you're bound to come across `ConsensusService` sooner rather than later by virtue of needing to inject it as a dependency, so these context-depending operations will become obvious sooner rather than later.
That said, we could raise an exception on every method call until `service.context` is no longer undefined — this guarantees the user is aware of the service's needs. @bithavoc, @MakotoTheKnight, is this, as well as other potential improvements to these context-dependent services, something we want to focus on now?

app/assets/javascripts/client/views/consensus/header.html
juliosantos: I don't agree. Conditionals are always harder to understand than a linear flow, and as such hiding them away or disguising them makes the code more readable. The reader of this piece of code doesn't have to think "ok, so if the context is an assessment, these paragraphs are rendered; if it's an analysis, these others are"; they think "ok, so some instruction paragraphs are rendered here".
That said, this opinion isn't too strong, and I am happy to make the change if you insist.

app/views/v1/analysis_consensus/show.json.jbuilder
juliosantos: Probably not, since a) it works and b) it's how it's been done elsewhere, in `app/views/v1/report/consensus_report.json.jbuilder`, `app/views/v1/responses/show.json.jbuilder` and `app/views/v1/responses/show_slimmed.json.jbuilder`.

It's my first time using `jbuilder`, but if I had to guess what's happening here, I'd say sending `number` to `json` is caught by `method_missing` in order to return a method which defines a key in the resulting `JSON` with that very name, `number`. In turn, this returned method is sent `number += 1`, which evaluates to an integer (to be the value for this `number` key) and increments `number` for the next cycle.

app/assets/javascripts/client/inventories/analysis_consensus_answer.html
juliosantos: Interesting question. In my opinion, which would be humble if not endorsed by Douglas Crockford, one should always use the strict equality operator in JavaScript. By default. Every time.

That said, I really don't think it's worth the change here, because:

0. Both objects under comparison are not variables, but integers. You can't redefine the integer `0` to be the string `derp`, no matter what you do in JavaScript. This means neither `0` or `1` above will ever undergo any typecasting, save for a hypothetical sun flare.
0. `==` defers to `===` if both comparables are of the same type. You'd think this proxying would have harrowing performance implications but, as it turns out, it doesn't. Look at the code below. It sure does tell us that `===` takes a mere 99% of what `==` does. However, this 1% amounts to only approximately 0.02 seconds, and that's with a billion comparisons.

```
> console.time("==");
> for (var i=0; i<1000000000; i++) { 0 == 1; }
> console.timeEnd("==");
> console.time("===");
> for (var i=0; i<1000000000; i++) { 0 === 1; }
> console.timeEnd("===");
< ==: 2070.465ms
< ===: 2051.804ms

> console.time("==");
> for (var i=0; i<1000000000; i++) { 0 == 1; }
> console.timeEnd("==");
> console.time("===");
> for (var i=0; i<1000000000; i++) { 0 === 1; }
> console.timeEnd("===");
< ==: 2042.784ms
< ===: 1998.043ms

> console.time("==");
> for (var i=0; i<1000000000; i++) { 0 == 1; }
> console.timeEnd("==");
> console.time("===");
> for (var i=0; i<1000000000; i++) { 0 === 1; }
> console.timeEnd("===");
< ==: 2013.570ms
< ===: 2003.153ms

> ((2051.804+1998.043+2003.153)/(2070.465+2042.784+2013.570))
< 0.9879514965269907
```

juliosantos:
@bithavoc Merged staging into this (rebasing this one would have been way harder as I saw the other day)

app/assets/javascripts/client/inventories/analysis_consensus_answer.html
juliosantos: But of course, what we should really be using is probably `<div ng-if="false">`

app/assets/javascripts/client/inventories/analysis_consensus_answer.html
```diff
@@ -0,0 +1,58 @@
...
+  </div>
+  <div class="score">
+    <div ng-repeat="answer in question.answers">
+      <span class="score-title">{{analysisConsensusAnswer.answerTitle(answer.value)}}</span>
+      <label class="score-checkbox scored-{{answer.value}}">
+        <fake-radio proxy-ng-model="question.score.value" proxy-ng-value="answer.value"></fake-radio>
+      </label>
+    </div>
+    <!-- Temporary hack for hiding "skip". -->
+    <div ng-if="false">
```
juliosantos: @bithavoc here :)

app/assets/javascripts/client/consensus/ConsensusService.js
MakotoTheKnight: @juliosantos:  Not an immediate focus, but good to keep in the back of our minds.

MakotoTheKnight:
@juliosantos:  I've done two things:

 - Merged the latest staging work in
 - Added `:id` to the list of accepted params for supporting_inventory_questions, as I observed new scores being created and old ones' score_id being nulled out.

MakotoTheKnight:
RE https://github.com/MobilityLabs/pdredesign-server/pull/306#issuecomment-218518171:  This is a papercut, but isn't sufficient enough to prevent us from releasing the code.  As far as I can tell the core functionality is there but we do need to tighten up the flow.

MakotoTheKnight:
I didn't see anything in this PR to cause me major concerns with it (outside of what I've already pushed), so this looks good to me.  :shipit:

We'll be merging this in immediately since we want to deploy this to staging today, and production Monday.

PR #304, opened 2016-05-09
Closes #229: Revised home page

MakotoTheKnight:
This PR is good to go.  Notes:

 - This updates the create buttons used on the site by giving them a more generic style and provides generic functionality for them.
 - The links to various things, such as the list of all analyses and the FAQ pages for each element do not presently work.  Work from #156 will dovetail with this (with the caveat that the routes in that task do not change from what they presently are).
 - This changes how we close a handful of modals - specifically, this changes how we close the creation modals for inventories, analyses, and assessments.  Instead of passing events up the chain we inject `$modalInstance`.  This has also led to us using the `resolve` block to pass in an `Inventory` when creating an analysis.

I've run through as many cases as I can but I'd like these to be smoketested:
   - Facilitator creating any one of the tools
   - Network Partner recommending any one of the tools
   - Participant creating any one of the tools

 Three accounts in each of those states should be created to test.  The behavior in terms of text and appearance of text should follow that of the Readiness Assessment.

PR #309, opened 2016-05-11
Closes #182: about pages

### LINKED ISSUE #182: About pages
Waiting for #170

juliosantos:
This PR is blocked waiting for #304 because of the reshuffling of modals.

Notes:

0. FAQs for Inventory and Analysis don't exist, so neither do their links on the sidebar
0. Sidebar isn't sliced up in parts, as it was already brought up elsewhere by @bithavoc, because of underlying CSS


MakotoTheKnight:
You are unblocked! :)  Go ahead and rebase against staging.

juliosantos:
@MakotoTheKnight done! It's now ready for review.

app/assets/javascripts/client/about/AboutSidebarCtrl.js
```diff
@@ -0,0 +1,68 @@
...
+    '$element',
+    '$location',
+    '$timeout',
+    '$modal'
+  ];
+
+  function AboutSidebarCtrl($scope, $rootScope, $element, $location, $timeout, $modal) {
+    var vm = this;
+
+    new Clipboard("[data-clipboard-target]");
```
MakotoTheKnight: Strange...I'm sure this is how Clipboard is initialized but it's a bit peculiar to me.

app/assets/javascripts/client/about/AboutSidebarCtrl.js
```diff
@@ -0,0 +1,68 @@
...
+
+    vm.openAssessmentModal = function() {
+      $modal.open({
```
MakotoTheKnight: So this applies to all of the modals:

 - What happens when I have more than one type open?  Is it the case that the other modal is guaranteed to be destroyed?
 - What happens when an unauthenticated user invokes these?  Should we notify them that they have to sign in to have the full functionality of it?

app/assets/javascripts/client/about/about_sidebar.html
```diff
@@ -0,0 +1,32 @@
...
+  <div class="sidebar-section">
+    <h2>PDredesign</h2>
+    <p><a ui-sref="about.pdredesign">About PDredesign</a></p>
+  </div>
+  <div class="sidebar-section">
+    <h2>1. Readiness Assessment</h2>
+    <p><a ui-sref="about.ra-assessment">About Readiness Assessment</a></p>
+    <p><a ui-sref="faqs({topic: 'Readiness Assessment'})">FAQs</a></p>
+    <p><a ng-click="aboutSidebar.openAssessmentModal()">Facilitate New Assessment</a></p>
+    <p><a ui-sref="assessments">View Assessment List</a></p>
```
MakotoTheKnight: This applies to all of the links in the sidebar:  what should I get when I click on a link and I'm not authenticated?

app/assets/javascripts/client/about/dt-analysis.html
```diff
@@ -0,0 +1,63 @@
+<div class="about-header">
+  <h1>Data & Tech Landscape - Analysis</h1>
```
MakotoTheKnight: Nitpick:  I'd prefer `&amp;` instead of `&` where necessary.

MakotoTheKnight:
Copy looks good - Didn't spot any issues with that and the transitions between the two phases are working well.  Other comments are in source code.

app/assets/javascripts/client/about/AboutSidebarCtrl.js
```diff
@@ -0,0 +1,68 @@
...
+  AboutSidebarCtrl.$inject = [
+    '$scope',
+    '$rootScope',
+    '$element',
+    '$location',
+    '$timeout',
+    '$modal'
+  ];
+
+  function AboutSidebarCtrl($scope, $rootScope, $element, $location, $timeout, $modal) {
```
MakotoTheKnight: `$scope` is unused here

app/assets/javascripts/client/about/AboutSidebarCtrl.js
juliosantos: Is it the use of `new` or the odd selector that you find peculiar?

MakotoTheKnight: `new`; we're instantiating something then immediately letting it fall on the floor.  The selector I'm fine with; I expect that to be an attribute somewhere in the DOM which the clipboard is being bound to.

juliosantos: I can assign it to `vm` if you like, but I don't see what we'd use this for.

app/assets/javascripts/client/about/AboutSidebarCtrl.js
juliosantos: You can't open more than one modal at the same time. After opening a modal, attempting to click a trigger for another one will just close the first.

Very good question re authentication. I guess I'll defer to @candyj0311

app/assets/javascripts/client/about/about_sidebar.html
juliosantos: The current behaviour is that you're redirected to the login page, presumably because of `authenticate: true` in the routes. I didn't change any of that.

MakotoTheKnight:
Functionality-wise, this looks fine.  There is an issue with the modals not behaving well when the user is unauthenticated, but the Product team has decided to deploy this to staging today.

PR #313, opened 2016-05-11
Closes #245: Landing page

### LINKED ISSUE #245: Updated Landing Page
This is what users see when they are not logged in. @HeyCat will attach the mockup once it's finished being updated.

MakotoTheKnight:

 - A lot of old images were removed; these images as best as I can tell were only being used on the home page.
 - A new directive was created to help with the testimonials.
 - This page is accessible to those who are unauthenticated.
 - The links should be pointing to those referred to by #182 and will be functional once its associated PR (#309) is merged in.


app/assets/javascripts/client/home/home_anon.html.erb
```diff
@@ -1,103 +1,66 @@
...
-    <div class='col-sm-12 text-center'>
-      <p>A digital toolkit and community of practice</br>
-        dedicated to improving professional</br>
-        development for teachers.</p>
-    </div>
+<div class="hero-header">
+  <div class="hero-banner">
+    <h1>PDredesign</h1>
+    <p class="text-center">A digital toolkit and community of practice dedicated to improving
+      <u>professional development</u> for teachers.</p>
```
juliosantos: Not a big deal at all, but I think non-semantic/style tags should play no role in HTML. I'd have used `<span class="underlined">` instead.

app/assets/javascripts/client/home/testimonal.js
```diff
@@ -0,0 +1,25 @@
+(function() {
+  'use strict';
+
+  angular.module('PDRClient')
+      .directive('testimonial', testimonial);
+
+  function testimonial() {
+    return {
+      restrict: 'E',
+      transclude: true,
```
juliosantos: Why the transclusion?

app/assets/javascripts/client/home/testimonal.js
```diff
@@ -0,0 +1,25 @@
...
+        image: '@',
+        backgroundStyle: '@',
+        learnMoreLink: '@',
+        quote: '@',
+        author: '@',
+        subtext: '@',
+        toolName: '@',
+        alternate: '@'
+      },
+      templateUrl: 'client/home/testimonial_template.html'
```
juliosantos: Isn't `_template` redundant on an HTML file?

app/assets/stylesheets/home/testimonial.sass
```diff
@@ -0,0 +1,52 @@
...
+    @extend %testimonial-column
+    .quote
+      border: 0
+      padding: 10px 20px
+      margin-bottom: 0
+      q
+        quotes: "\201C" "\201D" "\2018" "\2019"
+        font-size: 1.35em
+        &:before
+          content: open-quote
```
juliosantos: I didn't know about this and absolutely love it

juliosantos:
0. Just like throughout the application, there's a few visual inconsistencies which we decided we'd handle post-delivery. However, I thought I'd still point out how different the footer is with regards to the position of the "stay connected" icons.
0. <bikeshedding>Seeing the  logo everywhere feels weird.</bikeshedding>
0. IMO neither these or any of my other comments are showstoppers so LGTM

app/assets/javascripts/client/home/home_anon.html.erb
MakotoTheKnight: Fair point.

I did look for a CSS class since I don't particularly like the use of `<u>`, but this also follows suit from [Bootstrap's own recommendation](http://getbootstrap.com/css/#type-inline-text).

I can change it since I don't particularly agree with `<u>` anyway.

app/assets/javascripts/client/home/testimonal.js
MakotoTheKnight: I've noticed that this version of Angular will get very mixed up when compiling the assets if the name of the directive and the name of the template are similar enough, since it all gets compiled down to HTML.  I can experiment but I don't think that this is likely to change.

app/assets/javascripts/client/home/testimonal.js
MakotoTheKnight: Muscle memory, for the most part.  It doesn't look like I need it so I'll remove it.

juliosantos:
LGTM

PR #354, opened 2016-05-25
Closes #233: export analysis to csv

### LINKED ISSUE #233: Export Analysis Data to CSV
This should be very similar to what we use for the RA since the formats are similar.

juliosantos:


juliosantos:
Not into master.

PR #379, opened 2016-07-11
Closes #344: Inventory participants creating analyses

### LINKED ISSUE #344: Inventory participants should be able to create analyses on that inventory
The `InventoryAuthorizer` is set up to prevent this, but shouldn't be.

MakotoTheKnight:

 - Users which are listed as participants on an inventory may now create analyses from an inventory they're participating in, or from the top-level menu.

To verify:

 - Invite a user as a participant to an inventory.
 - Use that user to create an analysis as part of the inventory report page.
 - Use that user to create an analysis as part of the inventory listing page.

This will be deployed to staging and the ticket will be formally closed once verified.


MakotoTheKnight:
Closed due to wrong merge target.

